PR #6, opened 2019-06-11
Add support for GPIO AF14, bump version

Disasm:
This PR adds support for GPIO configuration with AF14, which is needed for USB peripheral.


dfrankland:
I was able to build for all supported devices. Thanks for the help @Disasm! 🎉

dfrankland:
Just published

Disasm:
Thanks!

PR #18, opened 2019-07-21
Support HSE and USB clock

agalakhov:


src/rcc.rs
```diff
@@ -221,50 +235,67 @@ impl CFGR {
...
 
+        // set prescalers and clock source
+        rcc.cfgr.modify(|_, w| unsafe {
+            w.ppre2()
+                .bits(ppre2_bits)
+                .ppre1()
+                .bits(ppre1_bits)
+                .hpre()
+                .bits(hpre_bits)
+                .usbpre()
```
dfrankland: Seems like the `stm32f301` and the `stm32f334` don't have a USB prescalers. If you'd like, I can fix this up with some conditional configuration.

dfrankland:
Thank you so much for contributing this! :smile:

src/rcc.rs
agalakhov: Thank you, I was not aware of this.

agalakhov: I added one more commit for feature-gating this.

dfrankland:
Awesome work, thanks again @agalakhov

PR #19, opened 2019-08-26
Update stm32f3 dependency, bump version

Disasm:


Disasm:
Checked with slightly modified script from `stm32f4xx-hal`

Disasm:
Thanks!

PR #24, opened 2019-11-13
Add USB driver and example

Disasm:
See also: https://github.com/stm32-rs/stm32-usbd/pull/16


Disasm:
Note that any of the `stm32f303x*` features implicitly activate `usbd-device` feature. It's not good in terms of the size of a dependency tree, but I don't know how to workaround this.

src/usb.rs
```diff
@@ -0,0 +1,47 @@
...
+
+use crate::stm32::{RCC, USB};
+use stm32_usbd::UsbPeripheral;
+
+use crate::gpio::gpioa::{PA11, PA12};
+use crate::gpio::AF14;
+pub use stm32_usbd::UsbBus;
+
+pub struct Peripheral {
+    pub usb: USB,
```
IamfromSpace: It seems like it’s more common (though I’m still fairly new to embedded rust) to keep structs opaque and then use “smart” constructor” functions that return them.

It seems like it might be worth doing so to keep the fields out of the public API, so it could be modified later.

dfrankland:
I was able to test these changes successfully with a simple keyboard example: https://github.com/dfrankland/proton-c/commit/7e1d54e1aa64efb810b5a162eda1a18d9aeda113

src/usb.rs
dfrankland: While that's true, this is also a very non-complicated struct that has properties publicly available through other interfaces. I think folks can expect breaking changes whilst we've yet to get to v1.0. A sister create `stm32f1xx-hal` has [already merged a similar struct](https://github.com/stm32-rs/stm32f1xx-hal/pull/135) so there's bit of precedent now to use something like this.

Disasm: Your approach is more compact in terms of code, but you have to keep in mind the order of the arguments. With this structure, you have named entities that you can set in any order.

IamfromSpace: Agreed on all counts, especially if there’s precedence already—more curiosity than anything else!

PR #25, opened 2019-11-13
Add StatefulOutputPin and ToggleableOutputPin

Sh3Rm4n:
I tested it on the STM32F303 / STM32DISCOVERY Board, with the method provided in the example.

### DIFF
--- a/examples/toggle.rs
+++ b/examples/toggle.rs
@@ -1,0 +1,41 @@
+//! Uses the StatefulOutputPin embedded_hal trait to toggle the pin
+//! On the stm32 discovery board this is the "south" led
+//! Target board: STM32F3DISCOVERY
+
+#![deny(unsafe_code)]
+#![no_main]
+#![no_std]
+
+extern crate panic_semihosting;
+
+use cortex_m_rt::entry;
+use stm32f3xx_hal::prelude::*;
+use stm32f3xx_hal::stm32;
+
+#[entry]
+fn main() -> ! {
+    let dp = stm32::Peripherals::take().unwrap();
+
+    let mut rcc = dp.RCC.constrain();
+    let mut gpioe = dp.GPIOE.split(&mut rcc.ahb);
+
+    let mut led = gpioe
+        .pe13
+        .into_push_pull_output(&mut gpioe.moder, &mut gpioe.otyper);
+
+    led.set_low().unwrap();
+
+    loop {
+        led.toggle().unwrap();
+        cortex_m::asm::delay(8_000_000);
+        // Toggle by hand.
+        // Uses `StatefulOutputPin` instead of `ToggleableOutputPin`.
+        // Logically it is the same.
+        if led.is_set_low().unwrap() {
+            led.set_high().unwrap();
+        } else {
+            led.set_low().unwrap();
+        }
+        cortex_m::asm::delay(8_000_000);
+    }
+}
--- a/src/prelude.rs
+++ b/src/prelude.rs
@@ -2,6 +2,14 @@
 
 pub use crate::flash::FlashExt as _stm32f3xx_hal_flash_FlashExt;
 pub use crate::gpio::GpioExt as _stm32f3xx_hal_gpio_GpioExt;
+#[cfg(feature = "unproven")]
+pub use crate::hal::digital::v2::OutputPin as _embedded_hal_digital_OutputPin;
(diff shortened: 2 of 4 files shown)


Sh3Rm4n:
NOTE: I shamelessly stole the implementation from https://github.com/stm32-rs/stm32f1xx-hal. 
As this is, as of my understanding, a generic method to read out the state of the pin, this should work for every stm32 gpio register.

dfrankland:
This looks great. I'll be looking to get this in once we test out and merge #21 since it has some hefty changes to `gpio.rs`.

examples/toggle.rs
```diff
@@ -0,0 +1,42 @@
...
+use stm32f3xx_hal::stm32;
+
+#[entry]
+fn main() -> ! {
+    let p = stm32::Peripherals::take().unwrap();
+
+    let mut rcc = p.RCC.constrain();
+    let mut gpioe = p.GPIOE.split(&mut rcc.ahb);
+
+    let mut led_red = gpioe
```
dfrankland: Might we be able to name this just `led`? The LED on my board is white :stuck_out_tongue_closed_eyes:

examples/toggle.rs
```diff
@@ -0,0 +1,42 @@
+//! Uses the StatefulOutputPin embedded_hal trait to toggle the pin
+//! On the stm32 discovery board this is the red led
+//! Target board: STM32F3DISCOVERY
+
+#![deny(unsafe_code)]
+#![no_main]
+#![no_std]
+
+use panic_semihosting as _;
```
dfrankland: To stay consistent with:
https://github.com/stm32-rs/stm32f3xx-hal/blob/902b56f8f11221cb8ffbf24305d66ecea858bf6c/examples/usb_serial.rs#L6

```suggestion
extern crate panic_semihosting;
```

src/prelude.rs
```diff
@@ -2,6 +2,8 @@
 
 pub use crate::flash::FlashExt as _stm32f3xx_hal_flash_FlashExt;
 pub use crate::gpio::GpioExt as _stm32f3xx_hal_gpio_GpioExt;
+pub use crate::hal::digital::v2::StatefulOutputPin as _embedded_hal_digital_StatefulOutputPin;
+pub use crate::hal::digital::v2::ToggleableOutputPin as _embedded_hal_digital_ToggleableOutputPin;
```
dfrankland: These re-exports needed to be guarded by `#[cfg(feature = "unproven")]`, right? Might be nice to re-export the `OutputPin` and `InputPin` traits too.

examples/toggle.rs
```diff
@@ -0,0 +1,42 @@
...
+
+use cortex_m_rt::entry;
+use stm32f3xx_hal::hal::digital::v2::OutputPin;
+use stm32f3xx_hal::prelude::*;
+use stm32f3xx_hal::stm32;
+
+#[entry]
+fn main() -> ! {
+    let p = stm32::Peripherals::take().unwrap();
```
dfrankland: To stay consistent with:
https://github.com/stm32-rs/stm32f3xx-hal/blob/902b56f8f11221cb8ffbf24305d66ecea858bf6c/examples/usb_serial.rs#L17

```suggestion
    let dp = stm32::Peripherals::take().unwrap();
```

examples/toggle.rs
Sh3Rm4n: Oh yeah right :grinning: I've not considered, that the led colors are not consistent between the discovery boards.

examples/toggle.rs
Sh3Rm4n: Personally, I am more in favor of the `use panic_semihosting as _;` method, as this is more Edition 2018 idiomatic (?)

However, consistency is more important, and I will happily change it.

src/prelude.rs
Sh3Rm4n: Yeah, you are right. I've considered it myself. Btw. is the cfg attribute always line based? Or is there another to apply it to a whole block / paragraph?

dfrankland: So far as I know, it's line based only. I've used [`cfg_if`](https://github.com/alexcrichton/cfg-if) with success for block based application.

Sh3Rm4n:
Applied change requests. Ready for review again :)

dfrankland:
Nice work 🎉

PR #26, opened 2019-11-15
Suppress dead code warning in gpio module.

IamfromSpace:
Allow dead code for the AFRL impl, and document the exact cases that require it.

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -252,6 +252,11 @@ macro_rules! gpio {
                 }
 
                 impl AFRL {
+                    // A couple device/port combos have no valid alternate functions:
+                    //   - stm32f303 GPIOG and GPIOH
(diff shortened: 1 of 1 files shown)


IamfromSpace:
Was able to figure out the exact problem cases, and I think this was the only source of warnings.  However, I don’t have access to my laptop currently, so I wasn’t able to run or validate this.  It’s a small change, but if someone were to give it a sanity check, that would probably be a good idea!

Otherwise, I’ll post again as soon as I’m able myself :)

dfrankland:
I can run a quick sanity on this. This is good for now, later we can probably add to the macro to conditionally `impl AFRL`.

dfrankland:
I was able to build all device features without any warnings

IamfromSpace:
Awesome, thanks for the double check!

PR #28, opened 2019-11-19
Add CI

Sh3Rm4n:
I do think, that any CI to test this crate could be useful. 

I used the build script from https://github.com/stm32-rs/stm32l4xx-hal and slightly modified it.

It is building the crate in any feature configuration and builds the examples as well.


dfrankland:
Looks great :tada:

PR #31, opened 2019-11-24
Support devices with 2-bit PLLSRC fields

russellmcc:
This allows us to use a 72mhz sysclk on stm32f303.  In the current
version, this is artificially locked to 64mhz by the hal (since the hal will never change pwm sources).

I also think this fixes a bug where we would never actually use the hse as the pll source.

### DIFF
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -342,9 +342,9 @@ impl CFGR {
         if let Some((pllmul_bits, pllsrc)) = pll_options {
             // enable PLL and wait for it to be ready
             rcc.cfgr
-                .write(|w| w.pllmul().bits(pllmul_bits).pllsrc().variant(pllsrc));
+                .modify(|_, w| w.pllmul().bits(pllmul_bits).pllsrc().variant(pllsrc));
 
-            rcc.cr.write(|w| w.pllon().set_bit());
+            rcc.cr.modify(|_, w| w.pllon().set_bit());
 
             while rcc.cr.read().pllrdy().bit_is_clear() {}
         }
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -189,7 +189,12 @@ impl CFGR {
     }
 
     /// Returns a tuple of the (pllsrclk frequency, pllmul, and pllsrc).
-    #[cfg(not(any(feature = "stm32f302", feature = "stm32f303")))]
+    #[cfg(not(any(
+        feature = "stm32f302",
+        feature = "stm32f303xd",
+        feature = "stm32f303xe",
+        feature = "stm32f398"
+    )))]
     fn calc_pll(&self) -> (u32, u32, rcc::cfgr::PLLSRCW) {
         let pllsrcclk = self.hse.unwrap_or(HSI / 2);
         let pllmul = self.sysclk.unwrap_or(pllsrcclk) / pllsrcclk;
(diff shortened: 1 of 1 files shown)


dfrankland:
Hmmm, this seems to prevent USB from working properly. I tested using this example here:
https://github.com/dfrankland/proton-c/tree/master/examples/simple-keyboard

russellmcc:
Thanks for checking on this.  I'll try to check this out in the next couple days.

russellmcc:
I don't have a good test set-up for USB at the moment.  Could you help debug?  

Do you know if you're hitting asserts, or things appear to go through and then USB doesn't work?  I'm trying to narrow down the issue.

If you're not hitting asserts, could you try reverting the line

```
            rcc.cfgr
                .write(|w| w.pllmul().bits(pllmul_bits).pllsrc().variant(pllsrc));
```

to the previous

```
            rcc.cfgr.write(|w| w.pllmul().bits(pllmul_bits));
```

If that makes USB work again, that would be good info for me to go on.

dfrankland:
Yep, that was on the money. The new code adding `.pllsrc().variant(pllsrc)` would make the loop `while rcc.cr.read().hserdy().bit_is_clear() {}` never end. The old code still works.

Unfortunately, this works to set the clocks, but still prevents USB working properly.

russellmcc:
sorry, want to double check - do you mean the `while rcc.cr.read().pllrdy().bit_is_clear() {}` loop?  I don't see how the pll source could affect the `hserdry` flag since we set it after looping for `hserdy`.  Thanks for your time with this.

dfrankland:
Yeah, it's quite odd.

I pushed the debug logs I created to a branch here for inspection:
https://github.com/dfrankland/stm32f3xx-hal/blob/b2457a4d6c405310b67879cde7e3dfc6f005d45c/src/rcc.rs#L377-L379

Here's the logs I get:
```
[stm32f3xx-hal/src/rcc.rs:287] hclk <= 72_000_000 = true
[stm32f3xx-hal/src/rcc.rs:306] pclk1 <= 36_000_000 = true
[stm32f3xx-hal/src/rcc.rs:325] pclk2 <= 72_000_000 = true
[stm32f3xx-hal/src/rcc.rs:352] (&usb_ok, &usbpre, &usbclk_valid) = (
    true,
    true,
    true,
)
[stm32f3xx-hal/src/rcc.rs:356] "try to enable hse" = "try to enable hse"
[stm32f3xx-hal/src/rcc.rs:363] "start waiting on hserdy" = "start waiting on hserdy"
[stm32f3xx-hal/src/rcc.rs:365] "hserdy is... ready" = "hserdy is... ready"
[stm32f3xx-hal/src/rcc.rs:368] "try to enable PLL" = "try to enable PLL"
[stm32f3xx-hal/src/rcc.rs:377] "start waiting on pllrdy" = "start waiting on pllrdy"
```

This the call that invokes this code path:
https://github.com/dfrankland/proton-c/blob/14fe346b3219dd7e77b1c68b08f7bf65f6bc4573/examples/simple-keyboard/main.rs#L43-L49

dfrankland:
OH! I just noticed I made a typo :man_facepalming:. Yes, you are correct. I meant the `pllrdy` loop.

russellmcc:
Appreciate your help with this - looks like the issue was just an existing typo where we were overwriting the HSE configuration while setting the PLL configuration (`write` vs the intended `modify`).   This wasn't a problem before this PR because there was a secondary bug where we never used HSE as the PLL source, even if it was intended.  I also fixed some of the math with PLL multiplication.

Let me know if the latest changes work for you!  I was at least able to start up a pll based on the hse, but I still didn't test USB specifically.

dfrankland:
I was able to test that all device features compile and that USB works! :tada:

PR #32, opened 2019-11-24
Correct gpio alternate function tables

russellmcc:
The current ones don't take into account the differences between the
303xd variants and the 303xb variants.  It looks like the existing tables
came from the xb datasheets.  This means that when using the xd/xe
variants, some functionality is available that shouldn't be, and some
functionality is missing.


strom-und-spiele:
Before merging the pin config of the STM32F303x6/x8 should be added.

russellmcc:
I believe I've added the x6/x8 pin config.

One subtlety here is backwards compatibility, since previously selecting "stm32f303" would give you access to, e.g., the whole E port, but now you don't get access to it until you select a specific chip like "stm32f303xd" which has the E port.  I think this is okay since the E port doesn't work on all stm32f303 chips, but just wanted to mention it.

russellmcc:
FYI - still working on this.  Trying to make it so that the "stm32f303" feature gives you the lowest common denominator of stm32f303 devices, and the more specific features give you access to the AFs available on those chips

russellmcc:
Okay, I think I've got it, thanks for patience.  I had to add a bit of an unsightly argument to the gpio macro to _prevent_ certain ports being enabled for certain devices.  I guess the other way to handle this would be to make stm32f303xe not imply the stm32f303 feature, but that seems like a worse solution to me.  Happy to adjust if others have different opinions or can think of a better way to handle this that I missed.

dfrankland:
There's many variants, so I am sure there's no elegant way to handle this :sweat_smile: I've tested this out with a blinky and a USB example, additionally all features build just fine.

PR #33, opened 2019-11-25
Add analog gpio mode

russellmcc:
This is necessary when you're using a DAC.

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -43,6 +43,9 @@ pub struct PushPull;
 /// Open drain output (type state)
 pub struct OpenDrain;
 
+/// Analog mode (type state)
+pub struct Analog;
+
 /// Alternate function 0 (type state)
 pub struct AF0;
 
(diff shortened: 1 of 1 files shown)


strom-und-spiele:
This is also needed for ADC, and I did (partly verbatim :) ) the same implementation:
https://github.com/strom-und-spiele/stm32f3xx-hal/blob/adc/src/gpio.rs
would suggest to also add pupdr to `into_analog`.
As stated in table 72 in the reference manual, only 0b00 (floating) is a valid
config  if the pin is set to analog mode.

russellmcc:
Thanks, you are of course correct.  I've added pupdr configuration.

dfrankland:
@strom-und-spiele, @russellmcc, I think y'all might be even better suited than myself to help run this repo. Would either of you be interested in being added as collaborators?

strom-und-spiele:
Thanks for the praise - I'm actually interested in being able to create PRs, but I'm still deep in the "aah - so this is how that works" stadium ;)

russellmcc:
Mostly I'm hesitant due to plenty of other commitments for my time, but if you don't mind a somewhat lackadaisical collaborator I'd be happy to help out - I just can't promise I'll be able to devote any significant time to it.

dfrankland:
@russellmcc I am looking for folks who know more about embedded development, since I am mostly a beginner myself. No commitments needed, just a review here and there. Although, you don't need to be a collaborator to review code 😁

PR #34, opened 2019-11-28
Pwm channels

### LINKED ISSUE #23: Support PWM Output
Supporting PWM output across devices/timers would allow us to support peripherals like servos.

The f3 supports multiple output channels (same period, possibly different duty cycle), and it supports multiple pin-out, making the `PwmPin` trait from `embedded-hal` a good way to go.  The [stm32f1xx-hal](https://docs.rs/stm32f1xx-hal/0.4.0/stm32f1xx_hal/pwm/index.html) has done a lot of good work here, however I propose a slightly different approach, where pin configuration is left to the channels that are returned.  This has the benefit of not causing an explosion combinatorics, makes for a flatter (and more easily supported macro), supports channels that can output to multiple pins, and the resulting interface is fairly intuitive to use.  We can use type states to ensure that only channels that _have_ pins can be enabled.

I have a WIP branch [here](https://github.com/IamfromSpace/stm32f3xx-hal/compare/gpio_afi_macro...IamfromSpace:pwm-channels), with most of the macros working.

The resulting usage would look like:

```rust
// other use statements omitted
use stm32f3xx_hal::pwm::tim8;
use embedded_hal::PwmPin;

#[entry]
fn main() -> ! {
    let dp = stm32f303::Peripherals::take().unwrap();

    let clocks = rcc.cfgr.freeze(&mut flash.acr);

    let mut gpioc = dp.GPIOC.split(&mut rcc.ahb);
    let pc8 = gpioc.pc8.into_af4(&mut gpioc.moder, &mut gpioc.afrh);

    let mut gpiob = dp.GPIOB.split(&mut rcc.ahb);
    let pb9 = gpiob.pb9.into_af10(&mut gpiob.moder, &mut gpiob.afrh);

    let (_, _, ch3_without_pins, _) = tim8(
        dp.TIM8,
        256, // resolution of duty cycle
        50, // frequency of period
        &clocks, // To get the timer's clock speed
    );

    // None of these lines compile due to type states
    // let max_duty = ch3_without_pins.get_max_duty();
    // ch3_without_pins.set_duty(max_duty / 4);
    // ch3_without_pins.enable();

    let mut ch3 = ch3_without_pins
        .output_to_pb9(pb9)
        .output_to_pc8(pc8);

    let max_duty = ch3.get_max_duty();
    ch3.set_duty(max_duty / 4);
    ch3.enable();

    loop {}
}
```

IamfromSpace:
This supports 95% of all device/channel/pin combinations.  Those that are missing seem to not yet have register support from the underlying stm32f3 crate and mostly just stm32f373 is affected.

As noted from the previous issue, this interface is a bit different from the implementation in the f1xx-hal.  The key thing is that the pins are provided _after_ the channels are returned.

In the f1xx approach there's a combinatoric explosion of possible pin combos.  Just counting common pins for TIM2 there are nearly 2^11 pin combos--so a macro approach is intractable.  Instead, the API enables a bring-your-own-impl for any desired pin combo working.  This a clever approach and can often be quite useful, but it's also overhead for a user.  This alternative approach "just works."

There are also has numerous guard rails in place while remaining intuitive.  Timers/channels/pins not supported by a device cannot be accessed. Channels cannot be enabled until they have pins.  Invalid pins cannot be supplied to a channel. Pins cannot be modified or shared after usage.  A channel may use regular or complementary pins but not both.

When using type inference, the mechanisms behind these invariants is fairly invisible.

This is a large PR, so I'm happy to take the time to address any and all feedback.  I mostly wanted to show that this approach could in fact cover all cases.

Closes #23


IamfromSpace:
After merging in the most recent PR for f303 features, it now fails to build (all other devices work).  Error is that port D/E/F aren't available, but that doesn't quite make sense to me.  Hoping someone might have better insight as to how the previous changes might have affected this!

strom-und-spiele:
This is due to different functionalities on the different sub-versions of the f303. 
@russellmcc  wrote:
> One subtlety here is backwards compatibility, since previously selecting "stm32f303" would give you access to, e.g., the whole E port, but now you don't get access to it until you select a specific chip like "stm32f303xd" which has the E port. I think this is okay since the E port doesn't work on all stm32f303 chips, but just wanted to mention it.

see: https://github.com/stm32-rs/stm32f3xx-hal/pull/32

IamfromSpace:
@strom-und-spiele Ah, thanks for the clarification.  Was looking in the wrong spots before!  I'm a bit dubious about the ability to support sub-devices, but I see how the USB features don't leave any other option, so this makes sense.  Fixed now!

strom-und-spiele:
@IamfromSpace hth,
I'm not so sure on how to approach this issue with sub-devices either.
You looking around in the wrong places makes me suggest to really break backwards compatibility and force the usage of the correct subversion. After all (when using this crate), this is a thing you figure out once, put it in your Cargo file and never think about again. And imho it's not harder to determine the full version than just the main one. 

The crass alternative (not supporting sub-devices) is something I think would hold back the development of this crate.

dfrankland:
I am all up for breaking backwards compatibility. This crate is no where near `1.0.0` so I think it's natural to make some changes for the better.

IamfromSpace:
@strom-und-spiele @dfrankland agreed all around.  Supporting subdevices seems like a must, just quite a challenge today.  Having been working with a the device difference a lot lately, I’ve got some crazy ideas I’ll open a separate issue for—assuming I can get it concrete enough.

For this PR these changes were simple enough and are definitely the right way to go!

src/pwm.rs
```diff
@@ -0,0 +1,1683 @@
...
+        /// Create one or more output channels from a TIM Peripheral
+        /// This function requires the maximum resolution of the duty cycle,
+        /// the period of the PWM signal and the frozen clock configuration.
+        ///
+        /// The resolution should be chosen to offer sufficient steps against
+        /// your target peripheral.  For example, a servo that can turn from
+        /// 0 degrees (2% duty cycle) to 180 degrees (4% duty cycle) might choose
+        /// a resolution of 9000.  This allows the servo to be set in increments
+        /// of exactly one degree.
+        pub fn $timx(tim: $TIMx, res: $res, freq: Hertz, clocks: &Clocks) -> ($(PwmChannel<$TIMx_CHy, NoPins>),+) {
```
dfrankland: Could we silence the warnings here?

```suggestion
        #[allow(unused_parens)]
        pub fn $timx(tim: $TIMx, res: $res, freq: Hertz, clocks: &Clocks) -> ($(PwmChannel<$TIMx_CHy, NoPins>),+) {
```

dfrankland:
Awesome, thank you!

src/pwm.rs
IamfromSpace: Yeah, no simple way to avoid the warning--this is part of what makes single channels not result in a tuple.

Issue #36, opened 2019-12-03
Justify `.write` use


strom-und-spiele:
After following a few discussions on issues, it occurred to me that using `.write` reliably produced unwanted behavior. 
The only way I know how to solve this is to grep for `.write` and justify it's uses by adding comments or correct the wrong usages. This is some what ugly and requires future commits to hold to this practice, but it'd emphasize the importance of paying attention here.

I'd be happy to hear a cleaner solution and am wiling to implement it ( in the long run though - It seems like my life wants to be busy right now ;) )
I guess it won't be too much work at the current state of the repo (`$ rg "\.write" src/* | wc -l   → 26`)

IamfromSpace:
I totally agree with this.  When new to the write/modify I was using write inappropriately and it caused me endless headaches.

It seems like we would want modify in basically every case anyway.  I believe that the only reasons to use write are implicit reset and optimization.  Probably we’d want to prefer explicitness and optimization should be done with care—so they should be justified when done.

On a broader scope, based on the API, I originally used “write” as if it were modify when learning.  Maybe that’s just me being dumb, but it seems I’m not the only one.  Maybe it should be named “writeOrReset” or something to clarify what it does and steer users to modify (deprecating write).  I understand that that goes far beyond this package :)

Sh3Rm4n:
Totally agree. I was confused by the `write` method as well. This ultimatly comes down to how svd to rust works and there seems to be some ideas, how to tackle this. https://github.com/rust-embedded/svd2rust/pull/258#issuecomment-494698256

`write_or_reset` and names alike, which avoid confusion, should be the way to go, to replace `write`. Maybe we should propose a change to [svd2rust](https://github.com/rust-embedded/svd2rust)?

strom-und-spiele:
Did it on my fork: https://github.com/strom-und-spiele/stm32f3xx-hal will push it as a PR here if given access

dfrankland:
Hmm? No access necessary to make PRs 👍 go ahead and make one

strom-und-spiele:
hmm I got
```
]$ git push -u origin/justify_write
fatal: 'origin/justify_write' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```
with my git config containing
```
[remote "origin"]
        url = https://strom-und-spiele@github.com/stm32-rs/stm32f3xx-hal.git
[...]

PR #38, opened 2019-12-08
Justify write

strom-und-spiele:
Tends towards #36, removing some side-effects.
might have fixed some bugs that where not triggered until now.
Note that some side effects could be useful for the code working as intended, however
> Explicit is better than implicit.

### DIFF
--- a/src/pwm.rs
+++ b/src/pwm.rs
@@ -451,10 +451,8 @@ macro_rules! pwm_timer_private {
             // Set the "resolution" of the duty cycle (ticks before restarting at 0)
             // Oddly this is unsafe for some timers and not others
             //
-            // FIXME: I guess this is due to some Timers not allowing values >16bit.
-            // So this needs to be checked according to the Timer type.
-            //
-            // NOTE(write): uses all bits of this register.
+            // NOTE(write): not all timers are documented in stm32f3, thus marked unsafe.
+            // This write uses all bits of this register so there are no unknown side effects.
             #[allow(unused_unsafe)]
             tim.arr.write(|w| unsafe {
                 w.arr().bits(res)
--- a/src/pwm.rs
+++ b/src/pwm.rs
@@ -446,10 +446,15 @@ macro_rules! pwm_timer_private {
             }
 
             // enable auto reload preloader
-            tim.cr1.write(|w| w.arpe().set_bit());
+            tim.cr1.modify(|w| w.arpe().set_bit());
 
             // Set the "resolution" of the duty cycle (ticks before restarting at 0)
             // Oddly this is unsafe for some timers and not others
+            //
+            // FIXME: I guess this is due to some Timers not allowing values >16bit.
+            // So this needs to be checked according to the Timer type.
+            //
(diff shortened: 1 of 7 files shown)


src/pwm.rs
```diff
@@ -446,10 +446,15 @@ macro_rules! pwm_timer_private {
...
-            tim.cr1.write(|w| w.arpe().set_bit());
+            tim.cr1.modify(|_, w| w.arpe().set_bit());
 
             // Set the "resolution" of the duty cycle (ticks before restarting at 0)
             // Oddly this is unsafe for some timers and not others
+            //
+            // FIXME: I guess this is due to some Timers not allowing values >16bit.
+            // So this needs to be checked according to the Timer type.
+            //
+            // NOTE(write): uses all bits of this register.
```
Sh3Rm4n: Actually i think this unsafe is there because not all timers are documented in
`stm32f3`. And when they are not,this registers are marked as unsafe, because no
one programmatically confirmed that writing to this register has no side effects.

Sh3Rm4n:
I'm personally unsure about this change. `modify` is more natural to
use than write, however this introduces additional "read from register and
store it's values" instructions. I don't know if there are optimized away when
using release, but I have not enough experience to judge this.

When initializing a peripheral (for example writing to the control register), imo using `write` is ok, because these registers weren't touched before.

Maybe, when resetting the controller, the old values in the peripheral registers are kept, and we need to cleanly write new values to it, hence the `write` invocation. But this is just a wild guess.

strom-und-spiele:
I do agree with your objections. In some scenarios, `write` is the better choice. 

However this PR does not try to abolish `write` usage. The reason for starting this was that using `write` let to bugs in the past. The goal of this PR is to make the code more explicit.  

To give some examples:
* When writing to a stateless register, only `write` makes sense. (See, e.g. here: https://github.com/stm32-rs/stm32f3xx-hal/pull/38/files#diff-73928a829df74085c37fe73a4f3484fcR393 )
* When initializing a peripheral with a specific control register, `write` makes sure no _old_ values linger around. To make it clear to the developer what we are doing, we use `write` and add a comment.
(maybe here: https://github.com/stm32-rs/stm32f3xx-hal/pull/38/files#diff-ab68adbc36799b6f9885fbc8c64852c8R200 ? )
* When using only part of a control register, we might end up accidently disabling all other functionality (or do we want it here: https://github.com/stm32-rs/stm32f3xx-hal/pull/38/files#diff-dde82bd7fc015253d6e2e7f289dec7baR167 ? )

So to sum it up: Let's use `write` whenever it makes sense but let people know that we thought about why it makes sense.

src/pwm.rs
Sh3Rm4n: Unsafe is here, because of the reason I mentioned above. See for example [TIM17](https://stm32.agg.io/rs/STM32F303.html#TIM17:ARR) which ARR registers are still not documented in stm32f3. 

However, justifying write does still make sense. Remove the FIXME and this PR is ready to be merged :)

PR #41, opened 2019-12-22
Add rustfmt

Sh3Rm4n:
As this crate is almost rustfmt compatible, why don't we add it to the CI?

Also, this moves the rustup invocations to `install` which travis provides.
This make no difference now, but if we add more ci jobs with `matrix`, rustup wouldn't be called for every job.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,18 @@
 language: rust
+
 rust:
   - stable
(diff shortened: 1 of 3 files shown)


dfrankland:
Awesome! Still getting around to your other updates, thanks for your patience! 😋

PR #42, opened 2019-12-22
Use gpio variants

Sh3Rm4n:
This PR upgrades to variant methods for the gpio module.

Doing this I had to extend the macro invocations significantly.
Also, I found no elegant way to apply this to the downgradable pins, as the pin register field access is determined via
"runtime".

https://github.com/rust-embedded/svd2rust/pull/400 could help with that.

Also, for the `stm32f3x8` and `stm32f373` series, this can not be applied right now, as they are missing a gpio documentation.

https://github.com/stm32-rs/stm32-rs/pull/321 will add these.

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -107,10 +107,11 @@ macro_rules! gpio {
             gpio_mapped_iorst: $iopxrst:ident,
             partially_erased_pin: $PXx:ident,
             pins: [
-                $($PXi:ident: ($pxi:ident, $i:expr, $MODE:ty, $AFR:ident, [
-                    $($AFi_common:ty: ($into_afi_common:ident, $afi_common:expr,),)*
+                $($PXi:ident: ($pxi:ident, $i:expr, $MODE:ty, $moderi:ident, $AFR:ident, $afri:ident,
+                  $bsi:ident, $bri:ident, $odri:ident, $idri:ident, $pupdri:ident, $oti:ident, [
+                    $($AFi_common:ty: ($into_afi_common:ident, $afi_common:ident,),)*
                 ], [
-                    $($AFi:ty: ($into_afi:ident, $afi:expr, [$($afi_devices:expr,)*],),)*
+                    $($AFi:ty: ($into_afi:ident, $afi:ident, [$($afi_devices:expr,)*],),)*
                 ]),)+
             ],
         },)+
@@ -440,19 +441,8 @@ macro_rules! gpio {
                                 moder: &mut MODER,
                                 afr: &mut $AFR,
                             ) -> $PXi<$AFi_common> {
-                                let offset = 2 * $i;
-
-                                // alternate function mode
-                                let mode = 0b10;
-                                moder.moder().modify(|r, w| unsafe {
-                                    w.bits((r.bits() & !(0b11 << offset)) | (mode << offset))
-                                });
-
-                                let offset = 4 * ($i % 8);
-                                afr.afr().modify(|r, w| unsafe {
-                                    w.bits((r.bits() & !(0b1111 << offset)) | ($afi_common << offset))
-                                });
-
+                                moder.moder().modify(|_, w| w.$moderi().alternate());
+                                afr.afr().modify(|_, w| w.$afri().$afi_common());
                                 $PXi { _mode: PhantomData }
                             }
                         )*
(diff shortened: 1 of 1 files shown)


Sh3Rm4n:
Now that stm32f3 v0.10 is released, this builds for all stm3f3xx targets again. 

This blows up the macros in the gpio module to a large extent, but the code itself should be simplified. 

I would keep it like that for now. When the [field array support](rust-embedded/svd2rust#400) is released under a new svd2rust version and stm32f3 will use it in the next release, this code could be simplified a lot. 

When this happens, i will open a follow up PR.

ra-kete:
Hey @Sh3Rm4n, could you rebase to resolve the conflicts?

src/gpio.rs
```diff
@@ -678,24 +618,24 @@ macro_rules! gpio {
                         type Error = Infallible;
 
                         fn is_high(&self) -> Result<bool, Self::Error> {
-                            Ok(!self.is_low()?)
+                            Ok(unsafe { (*$GPIOX::ptr()).idr.read().$idri().is_high()})
```
ra-kete, line 621: It's good practice to have for every use of `unsafe` a comment that explains why the code is safe. In this case you can just copy the one from `is_low`.

src/gpio.rs
```diff
@@ -678,24 +618,24 @@ macro_rules! gpio {
...
-                            Ok(unsafe { (*$GPIOX::ptr()).idr.read().bits() & (1 << $i) == 0 })
+                            Ok(unsafe { (*$GPIOX::ptr()).idr.read().$idri().is_low()})
                         }
                     }
 
                     #[cfg(feature = "unproven")]
                     impl<MODE> StatefulOutputPin for $PXi<Output<MODE>> {
                         fn is_set_high(&self) -> Result<bool, Self::Error> {
-                            self.is_set_low().map(|b| !b)
+                            Ok(unsafe { (*$GPIOX::ptr()).odr.read().$odri().is_high()})
```
ra-kete, line 633: Here too, an `unsafe` comment would be good.

src/gpio.rs
```diff
@@ -729,208 +669,208 @@ gpio!([
...
+            ], [
+                AF2: (into_af2, af2, ["stm32f373", "stm32f378",],),
+                AF5: (into_af5, af5, ["stm32f301", "stm32f303xd", "stm32f303xe", "stm32f318", "stm32f398",],),
+                AF9: (into_af9, af9, ["stm32f302", "stm32f303", "stm32f373", "stm32f378", "stm32f334", "stm32f328", "stm32f358", "stm32f398",],),
+                AF10: (into_af10, af10, ["stm32f302", "stm32f303xb", "stm32f303xc", "stm32f303xd", "stm32f303xe", "stm32f373", "stm32f378", "stm32f358", "stm32f398",],),
+                AF11: (into_af11, af11, ["stm32f301", "stm32f318", "stm32f302", "stm32f303", "stm32f334", "stm32f328", "stm32f358", "stm32f398",],),
+                AF13: (into_af13, af13, ["stm32f334",],),
+                AF14: (into_af14, af14, ["stm32f302", "stm32f303xb", "stm32f303xc",],),
+            ]),
+            PA13: (pa13, 13, Input<Floating>, moder13, AFRH, afrh13, bs13, br13, odr13, idr13, pupdr13, ot13, [
```
ra-kete, line 837: The default mode of PA13, PA14, PA15, PB3 and PB4 is `AF0`, not `Input<Floating>`.
See #82

Sh3Rm4n: Ah this seems like a rebase mistake. Will fix it

src/gpio.rs
```diff
@@ -440,18 +441,10 @@ macro_rules! gpio {
                                 moder: &mut MODER,
                                 afr: &mut $AFR,
                             ) -> $PXi<$AFi_common> {
-                                let offset = 2 * $i;
-
                                 // alternate function mode
```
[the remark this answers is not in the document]
Sh3Rm4n: Yeah you are right. I did this patch for readability and more self-describing code in the first place, so these comments do not add any useful information anymore.

src/gpio.rs
```diff
@@ -575,18 +535,12 @@ macro_rules! gpio {
...
-                            moder.moder().modify(|r, w| unsafe {
-                                w.bits((r.bits() & !(0b11 << offset)) | (mode << offset))
-                            });
+                            moder.moder().modify(|_, w| w.$moderi().output());
 
                             // push pull output
-                            otyper
-                                .otyper()
-                                .modify(|r, w| unsafe { w.bits(r.bits() & !(0b1 << $i)) });
+                            otyper.otyper().modify(|_, w| w.$oti().open_drain());
```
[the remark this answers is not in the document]
Sh3Rm4n: Good catch! Yes, thanks!

src/gpio.rs
Sh3Rm4n: I discarded my change and kept the previous code. So no comment is needed and I do not add more `unsafe` keywords :grinning:

ra-kete:
Looks good, thanks!

PR #43, opened 2019-12-22
Add rcc variants

Sh3Rm4n:
This PR updates all register accesses to variants in the rcc and similar modules.

I took the opportunity to reorganize the usb rcc initialization to play nicer with variants.

~~I'm not really happy about:~~

https://github.com/stm32-rs/stm32f3xx-hal/blob/e09a86622c9dc9d421bd53f24f7ebc982505417a/src/rcc.rs#L292

~~but it works.~~

~~I wish [stm32rs](https://github.com/stm32-rs/stm32-rs) would implement `From<HPRE_A> -> u8`. I guess this is a [svd2rust](https://github.com/rust-embedded/svd2rust) thing.~~

Now using `u8::from()` for casting.

### DIFF
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -2,8 +2,10 @@
 
 use core::cmp;
 
-use crate::stm32::{rcc, RCC};
-use cast::u32;
+use crate::stm32::{
+    rcc::{self, cfgr},
+    RCC,
+};
 
 use crate::flash::ACR;
 use crate::time::Hertz;
@@ -195,14 +197,14 @@ impl CFGR {
         feature = "stm32f303xe",
         feature = "stm32f398"
     )))]
-    fn calc_pll(&self) -> (u32, u32, rcc::cfgr::PLLSRC_A) {
+    fn calc_pll(&self) -> (u32, u32, cfgr::PLLSRC_A) {
         let pllsrcclk = self.hse.unwrap_or(HSI / 2);
         let pllmul = self.sysclk.unwrap_or(pllsrcclk) / pllsrcclk;
 
         let pllsrc = if self.hse.is_some() {
-            rcc::cfgr::PLLSRC_A::HSE_DIV_PREDIV
+            cfgr::PLLSRC_A::HSE_DIV_PREDIV
         } else {
-            rcc::cfgr::PLLSRC_A::HSI_DIV2
+            cfgr::PLLSRC_A::HSI_DIV2
         };
         (pllsrcclk, pllmul, pllsrc)
     }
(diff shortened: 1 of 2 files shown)


src/rcc.rs
```diff
@@ -306,8 +306,13 @@ impl CFGR {
             })
             .unwrap_or(rcc::cfgr::HPRE_A::DIV1);
 
-        // TODO is this casting safe?
-        let hclk = sysclk / (1 << (hpre_div as u32));
+        let hpre_div_value = if u8::from(hpre_div) != 0 {
+            u8::from(hpre_div) - 7
+        } else {
+            0
+        };
```
dfrankland: May I suggest using [`TryFrom`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) / [`TryInto`](https://doc.rust-lang.org/std/convert/trait.TryInto.html)? This could be used for the other cases below as well.

Example ([link to playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2d95af577129c5c5cc3dbd2ac4a373a3)):
```rust
use std::convert::TryInto;

fn number_into_minus_seven<T: TryInto<u8>>(number: T) -> u8 {
    number.try_into().map(|x| x - 7).unwrap_or_default()
}

fn main() {
    assert_eq!(number_into_minus_seven(255), 248);
    assert_eq!(number_into_minus_seven(256), 0);
}
```

Sh3Rm4n: Thank you for the suggestion!

Sadly, this won't help. As the conversion from the enum to `u8` is not the
problem. The problem lies into subtracting the offset.

If I  understand it correctly, the conversion should always succeed, therefor
`try_from` is not needed.

I use `checked_sub` now. I could apply the tip using `unwrap_or_default`. Thank
you for that :)

dfrankland: Looks nice and clean :+1:

Sh3Rm4n:
Thank you for testing it on hardware, as I did not even test it myself yet :sweat_smile:

src/rcc.rs
```diff
@@ -214,91 +231,130 @@ impl CFGR {
...
-        let sysclk = pllmul * pllsrcclk;
+        let sysclk = pllsrcclk * pllmul;
         assert!(sysclk <= 72_000_000);
 
-        let pllmul_bits = pllmul as u8 - 2;
-        (sysclk, Some((pllmul_bits, pllsrc)))
+        // NOTE From<u8> for PLLMUL_A is not implemented
+        // https://docs.rs/stm32f3/0.9.0/src/stm32f3/stm32f303/rcc/cfgr.rs.html#629-662
+        // let pllmul_mul: cfgr::PLLMUL_A = (pllmul as u8 - 2).into();
+        let pllmul_mul = match pllmul as u8 {
```
Sh3Rm4n: I am looking to find a more elegant solution for this. After this I'm fine
merging this.

src/rcc.rs
```diff
@@ -214,91 +231,130 @@ impl CFGR {
...
+        // Get the shift factor by subtracting the offset
+        //
+        // `DIV2 - 1` is used for offset, as this is guaranteed to have only one value
+        // as opposed to `DIV1` which can be 0..=7
+        let hpre_div_factor = u8::from(hpre_div)
+            .checked_sub(u8::from(cfgr::HPRE_A::DIV2) - 1)
+            .unwrap_or_default();
 
-        let hclk = sysclk / (1 << (hpre_bits - 0b0111));
+        let hclk: u32 = sysclk / 1 << hpre_div_factor;
```
TimNN: I think there's a (pre-existing) bug in this line: Notice how `DIV32` does not exist (I also confirmed this in the reference manual). If I understand things correctly, that means that for `DIV64` and up the result here will be incorrect.

Sh3Rm4n: Good catch. Yes you are right. I'll try to fix this

src/rcc.rs
Sh3Rm4n: I've found no better solution to this, as we can't implement any `From` `TryFrom` etc. trait because of the orphan rule, which is okay, because this would reduce the amount of code. 

Maybe sometime this will get implemented in `svd2rust`.

src/rcc.rs
Sh3Rm4n: I've added a special case. If you want, you could take another look @TimNN

Sh3Rm4n: BTW the special case was missing before my change as well, so this should fix a bug which was there for some time.

Sh3Rm4n:
I'm now much more comfortable with the solution I used now. Now I do not try to awkwardly combining the meaning of the bit field values with the values used for calculation. 

I've copied the solution over from [stm32f4xx-hal](https://github.com/stm32-rs/stm32f4xx-hal/blob/master/src/rcc.rs#L263-L274). This should be more readable now.

Sh3Rm4n:
Hey @russellmcc, pinging you, as I know, that you have looked more closely into the rcc module :)

src/rcc.rs
```diff
@@ -214,92 +232,112 @@ impl CFGR {
...
+        (sysclk, Some((pllmul_mul, pllsrc)))
     }
 
     /// Freezes the clock configuration, making it effective
     pub fn freeze(self, acr: &mut ACR) -> Clocks {
         let (sysclk, pll_options) = self.calc_sysclk();
 
-        let hpre_bits = self
+        // TODO try this method
+        // https://github.com/stm32-rs/stm32f4xx-hal/blob/c888eaa4d7dcac9670102896e54945efb5c41a3b/src/rcc.rs#L263
```
russellmcc: Is this TODO still valid?

src/rcc.rs
```diff
@@ -214,92 +232,112 @@ impl CFGR {
...
+                40..=95 => (cfgr::HPRE_A::DIV64, 64),
+                96..=191 => (cfgr::HPRE_A::DIV128, 128),
+                192..=383 => (cfgr::HPRE_A::DIV256, 256),
+                _ => (cfgr::HPRE_A::DIV512, 512),
             })
-            .unwrap_or(0b0111);
+            .unwrap_or((cfgr::HPRE_A::DIV1, 1));
 
-        let hclk = sysclk / (1 << (hpre_bits - 0b0111));
+        let hclk: u32 = sysclk / 1 << hpre_div;
```
russellmcc: I don't understand this line.  I was expecting `let hclk = sysclk / hpre_div;`

Sh3Rm4n: You are right. The bitshift is not needed anymore, as the values, which where previously calculated from the bitshift, are explicitly listed now. I'll fix this.

src/rcc.rs
Sh3Rm4n: No :sweat_smile:

Sh3Rm4n:
So I gave it a short test on my discovery board and everything is working as expected. I just have to take a closer look and then this should be ready to be merged.

Sh3Rm4n:
So I finally tested it on hardware with some examples and it seems to be working :tada:

PR #47, opened 2019-12-29
Add ADC usage (for `stm32f303` devices)

### LINKED ISSUE #20: Support for ADCs
Hi,

I want to use the analog-to-digital converters on STM32F3. Since I need to implement this anyway, I think this HAL crate is the best place to add support.
* Can I just go ahead and submit pull requests when I have any?
* Otherwise, if someone is already working on this, is there any way I can help?
* I am assuming we need to implement Channel and OneShot from embedded_hal::adc. Is this correct?
* If not, are there any known problems with those traits that make them unsuitable for the SAR and ΣΔ converters available on STM32F3 chips?

My primary goal is using the ΣΔ ADCs on STM32F373, but I think implementing SAR and other STM32F3 chips may be reasonable because of a large amount of overlap. I would also love to hear any other thoughts about this.

Greetings,
Môshe van der Sterre

strom-und-spiele:
Close #20

### DIFF
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -1,0 +1,399 @@
+//! # Work in Progress
+//! API for the ADC1 and ADC2 (Analog to Digital Converter) on the 303xB/C/D/E
+//!
+//! IMPORTANT - only valid for the 303xB/C/D/E variant.
+//! NOT for the 303x6/8:
+//!
+//! The availability of ADC channels varies
+//! for different sub-variants of stm32f3xx controllers.
+//! This means that e.g. channel ADC1_IN11
+//! is connected to PB0 on a f303x6 chip
+//! while there is no such channel on a F303xB/C.
+//! However the PB0 on that variant is connected to channel ADC3_IN12.
+//!
+//! See → 15.3.4
+//!
+//!
+//! # Examples
+//! check `adc.rs` in the examples folder
+//!
+use cortex_m::asm;
+use embedded_hal::adc::{Channel, OneShot};
+
+use crate::{
+    gpio::{gpioa, gpiob, gpioc, gpiof, Analog},
+    rcc::{Clocks, AHB},
+    stm32::{ADC1, ADC1_2, ADC2},
+};
+
+/// ADC configuration
+pub struct Adc<ADC> {
+    rb: ADC,
+    clocks: Clocks,
+    prescale: Prescale,
+    operation_mode: Option<OperationMode>,
+}
+
+#[derive(Clone, Copy, Debug)]
+/// ADC sampling time
+///
+/// each channel can be sampled with a different sample time.
+/// the total conversion time is
+/// 12.5 ADC clock cycles + sample time (T_x + .5)
+///
+/// TODO: there are boundaries on how this can be set depending on the hardware.
+pub enum SampleTime {
+    T_1,
+    T_2,
+    T_4,
+    T_7,
+    T_19,
+    T_61,
+    T_181,
+    T_601,
+}
+
+impl SampleTime {
+    /// Get the default timer
+    pub fn default() -> Self {
+        SampleTime::T_19
+    }
+
+    /// Conversion to bits for SMP
+    fn bitcode(&self) -> u8 {
+        match self {
+            SampleTime::T_1 => 0b000,
+            SampleTime::T_2 => 0b001,
+            SampleTime::T_4 => 0b010,
+            SampleTime::T_7 => 0b011,
+            SampleTime::T_19 => 0b100,
+            SampleTime::T_61 => 0b101,
+            SampleTime::T_181 => 0b110,
+            SampleTime::T_601 => 0b111,
+        }
+    }
+}
+#[derive(Clone, Copy, Debug, PartialEq)]
+/// ADC operation mode
+pub enum OperationMode {
+    OneShot,
+    // TODO all the other modes
+}
+
+#[derive(Clone, Copy, Debug)]
+/// ADC prescale
+pub enum Prescale {
+    // HCLK_1 needs some requirements to be met
+    HCLK_2 = 2,
+    HCLK_4 = 4,
+}
+
+impl Prescale {
+    /// Get default prescaler
+    fn default() -> Self {
+        Prescale::HCLK_2
+    }
+
+    /// Conversion to bits for CKMODE in ADCx_CCR
+    fn bitcode(&self) -> u8 {
+        match self {
+            Prescale::HCLK_2 => 0b10,
+            Prescale::HCLK_4 => 0b11,
+        }
+    }
+}
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+/// ADC data register alignment
+pub enum Align {
+    /// Right alignment of output data
+    Right,
+    /// Left alignment of output data
+    Left,
+}
+
+impl Align {
+    /// Default: right alignment
+    pub fn default() -> Self {
+        Align::Right
+    }
+}
+
+impl Align {
+    /// Conversion to bits for ALIGN in ADCx_CFGR
+    fn bitvalue(&self) -> bool {
+        match self {
+            Align::Right => false,
+            Align::Left => true,
+        }
+    }
+}
+
+macro_rules! adc_pins {
+    ($ADC:ident, $($pin:ty => $chan:expr),+ $(,)*) => {
+        $(
+            impl Channel<$ADC> for $pin {
+                type ID = u8;
+
+                fn channel() -> u8 { $chan }
+            }
+        )+
+    };
+}
+
+#[cfg(any(feature = "stm32f303",))]
+adc_pins!(ADC1,
+    gpioa::PA0<Analog> => 1_u8,
+    gpioa::PA1<Analog> => 2_u8,
+    gpioa::PA2<Analog> => 3_u8,
+    gpioa::PA3<Analog> => 4_u8,
+    gpiof::PF4<Analog> => 5_u8,
+    // Channels 6 to 10 are shared channels (i.e. ADC12_INx)
+    gpioc::PC0<Analog> => 6_u8,
+    gpioc::PC1<Analog> => 7_u8,
+    gpioc::PC2<Analog> => 8_u8,
+    gpioc::PC3<Analog> => 9_u8,
+    gpiof::PF2<Analog> => 10_u8,
+);
+
+#[cfg(any(feature = "stm32f303",))]
+adc_pins!(ADC2,
+    gpioa::PA4<Analog> => 1_u8,
+    gpioa::PA5<Analog> => 2_u8,
+    gpioa::PA6<Analog> => 3_u8,
+    gpioa::PA7<Analog> => 4_u8,
+    gpioc::PC4<Analog> => 5_u8,
+    // Channels 6 to 10 are shared channels (i.e. ADC12_INx)
+    gpioc::PC0<Analog> => 6_u8,
+    gpioc::PC1<Analog> => 7_u8,
+    gpioc::PC2<Analog> => 8_u8,
+    gpioc::PC3<Analog> => 9_u8,
+    gpiof::PF2<Analog> => 10_u8,
+    // exclusive ADC2 channels again
+    gpioc::PC5<Analog> => 11_u8,
+    gpiob::PB2<Analog> => 12_u8,
+);
+
+macro_rules! adc_hal {
+    ($(
+            $ADC:ident: ($init:ident, $ADC_COMMON:ident),
+    )+) => {
+        $(
+            impl Adc<$ADC> {
+                /// Init a new ADC
+                ///
+                /// Enables the clock, performs a calibration and enables the ADC
+
+                pub fn $init(
+                    rb: $ADC,
+                    adc_common : &mut $ADC_COMMON,
+                    ahb: &mut AHB,
+                    clocks: Clocks,
+                ) -> Self {
+                    let mut this_adc = Self {
+                        rb,
+                        clocks,
+                        prescale : Prescale::default(),
+                        operation_mode: None,
+                    };
+                    this_adc.enable_clock(ahb, adc_common);
+                    this_adc.set_align(Align::default());
+                    this_adc.calibrate();
+                    // ADEN bit cannot be set during ADCAL=1 and 4 ADC clock cycle after the ADCAL
+                    // bit is cleared by hardware
+                    this_adc.wait_adc_clk_cycles(4);
+                    this_adc.enable();
+                    return this_adc;
+                }
+
+                /// sets up adc in one shot mode for a single channel
+                pub fn setup_oneshot(&mut self) {
+                    // stop and clear overrun events
+                    self.rb.cr.modify(|_, w| w.adstp().set_bit());
+                    self.rb.isr.modify(|_, w| w.ovr().clear_bit());
+
+                    self.rb.cfgr.modify(|_, w| w
+                        .cont()     .clear_bit()
+                        .ovrmod()   .clear_bit()
+                    );
+
+                    self.rb.sqr1.modify(|_, w|
+                        // NOTE(unsafe): set the sequence length to 1
+                        unsafe { w.l3().bits(0) }
+                    );
+
+                    self.operation_mode = Some(OperationMode::OneShot);
+                }
+
+                fn set_align(&self, align: Align) {
+                    self.rb.cfgr.modify(|_, w| w.align().bit(align.bitvalue()));
+                }
+
+                fn enable(&mut self) {
+                    self.rb.cr.modify(|_, w| w.aden().set_bit());
+                    while self.rb.isr.read().adrdy().bit_is_clear() {}
+                }
+
+                fn disable(&mut self) {
+                    self.rb.cr.modify(|_, w| w.aden().clear_bit());
+                }
+
+                fn enable_clock(&self, ahb: &mut AHB, adc_common: &mut $ADC_COMMON) {
+                    ahb.enr().modify(|_, w| w.adc12en().enabled());
+                    unsafe {
+                        adc_common.ccr.modify(|_, w| w
+                            .ckmode().bits(self.prescale.bitcode())
+                        );
+                    }
+                }
+
+                /// Calibrate according to 15.3.8 in the Reference Manual
+                fn calibrate(&mut self) {
+                    if !self.advregen_enabled() {
+                        self.advregen_enable();
+                        self.wait_advregen_startup();
+                    }
+
+                    self.disable();
+
+                    self.rb.cr.modify(|_, w| w
+                        // NOTE: needs to be adopted if implementing differential input
+                        .adcaldif().clear_bit()
+                        .adcal()   .set_bit());
+
+                    while self.rb.cr.read().adcal().bit_is_set() {}
+                }
+
+
+                fn wait_adc_clk_cycles(&self, cycles: u32) {
+                    let adc_clk_cycle = self.clocks.hclk().0 / (self.prescale as u32);
+                    asm::delay(adc_clk_cycle * cycles);
+                }
+
+
+                fn advregen_enabled(&self) -> bool {
+                    return self._get_new_advregen() == 0b01;
+                }
+
+                fn advregen_enable(&mut self){
+                    // need to go though 00 first
+                    self._set_new_advregen(0b00);
+                    self._set_new_advregen(0b01);
+                }
+
+                /// returns ADVREGEN[1:0]
+                /// (deeppwd got merged as high bit in advregen - see ref manual)
+                fn _get_new_advregen(&self) -> u8 {
+                    return
+                        (self.rb.cr.read().deeppwd().bit() as u8) << 1 |
+                        (self.rb.cr.read().advregen().bit() as u8);
+                }
+
+                /// sets ADVREGEN[1:0]
+                /// (deeppwd got merged as high bit in advregen - see ref manual)
+                fn _set_new_advregen(&mut self, val: u8) {
+                    self.rb.cr.modify(|_, w| { w
+                        .deeppwd().bit((val & 0x02) != 0)
+                            .advregen().bit((val & 0x01) != 0)
+                    });
+                }
+
+                #[cfg(any(feature = "stm32f303",))]
+                fn wait_advregen_startup(&self) {
+                    const MAX_STARTUP_TIME_US: u32 = 10;
+                    asm::delay(MAX_STARTUP_TIME_US / (self.clocks.sysclk().0 /1_000_000));
+                }
+
+                fn convert_one(&mut self, chan: u8) -> u16 {
+                    self.ensure_oneshot();
+                    self.set_chan_smps(chan, SampleTime::default());
+                    self.select_single_chan(chan);
+
+                    self.rb.cr.modify(|_, w| w.adstart().set_bit());
+                    while self.rb.isr.read().eos().bit_is_clear() {}
+                    return self.rb.dr.read().regular_data().bits();
+                }
+
+                fn ensure_oneshot(&mut self) {
+                    match self.operation_mode {
+                        Some(mode) =>
+                        {
+                            if mode != OperationMode::OneShot {
+                                self.setup_oneshot();
+                            }
+                        },
+                        _ => self.setup_oneshot(),
+                    };
+                }
+
+                fn select_single_chan(&self, chan: u8) {
+                    self.rb.sqr1.modify(|_, w|
+                        // NOTE(unsafe): set the ADC_INx
+                        unsafe { w.sq1().bits(chan) }
+                    );
+                }
+
+                // Note: only allowed when ADSTART = 0
+                fn set_chan_smps(&self, chan: u8, smp: SampleTime) {
+                    match chan {
+                        1 => self.rb.smpr1.modify(|_, w|
+                            unsafe {w.smp1().bits(smp.bitcode())}),
+                        2 => self.rb.smpr1.modify(|_, w|
+                            unsafe {w.smp2().bits(smp.bitcode())}),
+                        3 => self.rb.smpr1.modify(|_, w|
+                            unsafe {w.smp3().bits(smp.bitcode())}),
+                        4 => self.rb.smpr1.modify(|_, w|
+                            unsafe {w.smp4().bits(smp.bitcode())}),
+                        5 => self.rb.smpr1.modify(|_, w|
+                            unsafe {w.smp5().bits(smp.bitcode())}),
+                        6 => self.rb.smpr1.modify(|_, w|
+                            unsafe {w.smp6().bits(smp.bitcode())}),
+                        7 => self.rb.smpr1.modify(|_, w|
+                            unsafe {w.smp7().bits(smp.bitcode())}),
+                        8 => self.rb.smpr1.modify(|_, w|
+                            unsafe {w.smp8().bits(smp.bitcode())}),
+                        9 => self.rb.smpr1.modify(|_, w|
+                            unsafe {w.smp9().bits(smp.bitcode())}),
+                        11 => self.rb.smpr2.modify(|_, w|
+                            unsafe {w.smp10().bits(smp.bitcode())}),
+                        12 => self.rb.smpr2.modify(|_, w|
+                            unsafe {w.smp12().bits(smp.bitcode())}),
+                        13 => self.rb.smpr2.modify(|_, w|
+                            unsafe {w.smp13().bits(smp.bitcode())}),
+                        14 => self.rb.smpr2.modify(|_, w|
+                            unsafe {w.smp14().bits(smp.bitcode())}),
+                        15 => self.rb.smpr2.modify(|_, w|
+                            unsafe {w.smp15().bits(smp.bitcode())}),
+                        16 => self.rb.smpr2.modify(|_, w|
+                            unsafe {w.smp16().bits(smp.bitcode())}),
+                        17 => self.rb.smpr2.modify(|_, w|
+                            unsafe {w.smp17().bits(smp.bitcode())}),
+                        18 => self.rb.smpr2.modify(|_, w|
+                            unsafe {w.smp18().bits(smp.bitcode())}),
+                        _ => unreachable!(),
+                    };
+                }
+
+            }
+
+            impl<WORD, PIN> OneShot<$ADC, WORD, PIN> for Adc<$ADC>
+            where
+                WORD: From<u16>,
+                PIN: Channel<$ADC, ID = u8>,
+                {
+                    type Error = ();
+
+                    fn read(&mut self, _pin: &mut PIN) -> nb::Result<WORD, Self::Error> {
+                        let res = self.convert_one(PIN::channel());
+                        return Ok(res.into());
+                    }
+                }
+        )+
+    }
+}
+
+#[cfg(any(feature = "stm32f303",))]
+adc_hal! {
+    ADC1: (adc1, ADC1_2),
+    ADC2: (adc2, ADC1_2),
+}
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -100,16 +100,13 @@ pub enum OperationMode {
     OneShot,
 }
 
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, PartialEq)]
 /// ADC CKMODE
 ///
 /// TODO: Add ASYNCHRONOUS mode
-/// TODO: ADD SYNCDIV1 mode with a check if the AHB prescaler of the
-/// RCC is set to 1 (the duty cycle of the AHB clock must be 50% in this configuration).
-/// see RM0316 15.3.3
 pub enum CKMODE {
     // ASYNCHRONOUS = 0,
-    // SYNCDIV1 = 1
+    SYNCDIV1 = 1,
     SYNCDIV2 = 2,
     SYNCDIV4 = 4,
 }
@@ -126,7 +123,7 @@ impl From<CKMODE> for stm32f3::stm32f303::adc1_2::ccr::CKMODE_A {
     fn from(ckmode: CKMODE) -> Self {
         match ckmode {
             //CKMODE::ASYNCHRONOUS => CKMODE_A::ASYNCHRONOUS,
-            //CKMODE::SYNCDIV1 => CKMODE_A::SYNCDIV1,
+            CKMODE::SYNCDIV1 => CKMODE_A::SYNCDIV1,
             CKMODE::SYNCDIV2 => CKMODE_A::SYNCDIV2,
             CKMODE::SYNCDIV4 => CKMODE_A::SYNCDIV4,
         }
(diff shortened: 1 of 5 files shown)


src/adc.rs
```diff
@@ -0,0 +1,534 @@
...
+                    };
+                    this_adc.enable_clock(ahb, adc_common);
+                    this_adc.set_align(Align::default());
+                    this_adc.calibrate();
+                    // ADEN bit cannot be set during ADCAL=1
+                    // and 4 ADC clock cycle after the ADCAL
+                    // bit is cleared by hardware
+                    this_adc.wait_adc_clk_cycles(4);
+                    this_adc.enable();
+                    return this_adc;
```
Sh3Rm4n:
```suggestion
                    this_adc
```

Return is not needed at the end of a function.

Sh3Rm4n:
Sadly not much of the ADC registers are documented in [stm32-rs](https://github.com/stm32-rs/stm32-rs/). Maybe we could submit a PR, which improves the situation. This would allow using variants, which provide nice type safety.

src/adc.rs
```diff
@@ -0,0 +1,534 @@
...
+    gpiod::PD14<Analog> => 11_u8,
+);
+
+/// Abstract implementation of ADC functionality
+///
+/// Do not use directly. See adc12_hal for a applicable Macro.
+/// TODO: Extend/generalize beyond f303
+macro_rules! adc_hal {
+    ($(
+            $ADC:ident: ($init:ident, $ADC_COMMON:ident),
```
Sh3Rm4n:
```suggestion
            $ADC:ident: ($adcx:ident, $ADC_COMMON:ident),
```
This would be more consistent with other macro definitions, see:

https://github.com/stm32-rs/stm32f3xx-hal/blob/7e0d7e67617fdb12174f76872a00bd2f54a6eb01/src/gpio.rs#L103

strom-und-spiele:
Thanks for the review I'll try to patch it the next days.

src/adc.rs
```diff
@@ -0,0 +1,534 @@
...
+}
+
+impl Prescale {
+    /// Get default prescaler
+    fn default() -> Self {
+        Prescale::HCLK_2
+    }
+
+    /// Conversion to bits for CKMODE in ADCx_CCR
+    fn bitcode(&self) -> u8 {
```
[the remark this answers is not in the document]
strom-und-spiele: Right now:
``` 
assert_eq!(4, <u8>::From(Prescale::HCLK_4));  # b.c. pub enum Prescale { […]￼    HCLK_4 = 4, ￼}
assert_eq!(3, Prescale::HCLK_4.bitcode());  # b.c. 0b11 <-> HCLK_4
```
This is why I also used the function `bitcode` in other situations. 

Any suggestions on how we can resolve this consistently?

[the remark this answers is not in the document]
strom-und-spiele: Actually, this originated in me using fmt::Debug and doesn't make sense anymore. Implementing into is probably the best idea.

src/adc.rs
```diff
@@ -0,0 +1,534 @@
...
+}
+
+impl SampleTime {
+    /// Get the default timer
+    pub fn default() -> Self {
+        SampleTime::T_19
+    }
+
+    /// Conversion to bits for SMP
+    fn bitcode(&self) -> u8 {
```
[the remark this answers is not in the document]
strom-und-spiele: [See context](https://github.com/stm32-rs/stm32f3xx-hal/pull/47#discussion_r363331439) :arrow_down:

src/adc.rs
```diff
@@ -0,0 +1,534 @@
...
+impl Align {
+    /// Default: right alignment
+    pub fn default() -> Self {
+        Align::Right
+    }
+}
+
+impl Align {
+    /// Conversion to bits for ALIGN in ADCx_CFGR
+    fn bitvalue(&self) -> bool {
```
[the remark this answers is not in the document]
strom-und-spiele: [See context](https://github.com/stm32-rs/stm32f3xx-hal/pull/47#discussion_r363331439) :arrow_up:

strom-und-spiele:
@Sh3Rm4n wrote:
> Sadly not much of the ADC registers are documented in [stm32-rs](https://github.com/stm32-rs/stm32-rs/). Maybe we could submit a PR, which improves the situation. This would allow using variants, which provide nice type safety.

I was under the impression that the variant usage here is a type safe way to discriminate between say the stm32f303xB/C/D/E and the stm32f303x6/8 variants. 
Or are you suggesting to add the discriminating definitions also to the stm32-rs package? (for as far as I understand it, the different handling of the variants is still needed here.)

Sh3Rm4n:
In this case, `stm32-rs` does not make a distinction between stm32f303xB/C/D/E and the stm32f303x6/8, as it's code is generated from an svd file for the stm32f303 family, IIRC.

I meant, that I wished more ADC peripherals / registers are documented, so we could use the "variant" methods for the ADC registers. This makes the code more clean and makes the intention clearer, when manipulating the register values. And it provides more type safety.

The ADC registers are mostly documented, as we can see [here](https://stm32.agg.io/rs/STM32F303.html).

~~But it's not worth the effort to use the variants, when there are not documented for the other stm32f3xx devices, yet.~~

... I justed noticed, that every ADC register is already documented? I don't remeber if this was already the case, when I last looked. Then this should allow us to use variant for all stm32f3xx controllers.

So there is no need to make an PR at `stm32-rs`, as this is already done.

I'm pretty busy right now, however I can have a look at this PR sometime this
weekend. :)

src/adc.rs
```diff
@@ -0,0 +1,535 @@
...
+                }
+
+                /// busy ADC read
+                fn convert_one(&mut self, chan: u8) -> u16 {
+                    self.ensure_oneshot();
+                    self.set_chan_smps(chan, SampleTime::default());
+                    self.select_single_chan(chan);
+
+                    self.rb.cr.modify(|_, w| w.adstart().set_bit());
+                    while self.rb.isr.read().eos().bit_is_clear() {}
```
ra-kete: The reference (15.3.24) mentions that the "EOS flag is cleared by the software by writing 1 to it". There doesn't seem to be another mechanism to clear it, so I think we need to clear it here to ensure the next read works as expected.

src/lib.rs
```diff
@@ -3,17 +3,10 @@
...
-        stm32f318
-        stm32f302
-        stm32f303
-        stm32f373
-        stm32f378
-        stm32f334
-        stm32f328
-        stm32f358
-        stm32f398"
+    "This crate requires you to specify your target device as a feautre.
```
ra-kete: typo: "feature"

src/adc.rs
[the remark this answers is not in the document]
Sh3Rm4n: Whoops yeah, you are right on this one. However, on `SampleTime` shouldn't #[repr(u8)] be sufficient? And then implementing the [Default](https://doc.rust-lang.org/std/default/trait.Default.html) trait by hand? 

For me this seems more idiomatic, but I have not a lot of experience regarding rust, especially with idiomatic rust.

Sh3Rm4n:
With the release of stm32f3 - v0.10.0, we could now take advantage of the type systems, as ADC is now fully documented! :) [`chkmode`](https://docs.rs/stm32f3/0.10.0/stm32f3/stm32f303/adc1_2/ccr/enum.CKMODE_A.html) for example

src/adc.rs
```diff
@@ -0,0 +1,535 @@
...
+/// TODO: Implement other modes (DMA, Differential,…)
+pub enum OperationMode {
+    OneShot,
+}
+
+#[derive(Clone, Copy)]
+/// ADC prescale
+///
+/// TODO: Implement a setter that checks requirements for `HCLK_1` and add it.
+pub enum Prescale {
```
Sh3Rm4n: As an example for stm32f3 v0.10.0 this enum could be replaced with [CKMODE_W](https://docs.rs/stm32f3/0.10.0/src/stm32f3/stm32f303/adc1_2/ccr.rs.html#328-337).

Sh3Rm4n:
Updated PR comment to link to issue directly :)

Sh3Rm4n:
Reminder (for later doesn't have to be implemented in this PR) 

The current implementation does not consider using the otther ADC prescaler: (the one below)

img-2020-03-12-232314

This is a bit more complicated to implement. On the [stm32f303b/c](https://www.st.com/content/ccc/resource/technical/document/reference_manual/4a/19/6e/18/9d/92/43/32/DM00043574.pdf/files/DM00043574.pdf/jcr:content/translations/en.DM00043574.pdf#page=126) it does use PLLCLK, which does seem more common. On the other hand the [stm32f303d/e](https://www.st.com/content/ccc/resource/technical/document/reference_manual/4a/19/6e/18/9d/92/43/32/DM00043574.pdf/files/DM00043574.pdf/jcr:content/translations/en.DM00043574.pdf#page=127) does use the HSI directly. 

Hoowever, inspecting this more carefully, this seems like a drawing error, as the Clock Mux on the stm32f303d/e has an HSI input, where nothing is connected to.

Sh3Rm4n:
> I was under the impression that the variant usage here is a type safe way to discriminate between say the stm32f303xB/C/D/E and the stm32f303x6/8 variants.

Unfortunately these cannot be distinguished, as these variants are all generated from the same peripheral description file (svd) file via svd2rust, which stm32f3 uses.

src/adc.rs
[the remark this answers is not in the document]
strom-und-spiele: I don't fully get your intentions. Let's dissect:
###  `#[repr(u8)]` 
As far as I understand it, this is only affecting the [Type Layout](https://doc.rust-lang.org/reference/type-layout.html#primitive-representations), i.e. we disallow the compiler to find a fitting size for each `SampleTime` object and fix the size to 8 bit. 
Of course in this scenario we'll never need more than 8 bit to discriminate between all the (fixed) variants, however I don't see why we should tell the compiler explicitly. This just creates another LOC without adding any functionality (that I can think of right now).

### implementing the Default by hand
[This is already implemented.](https://github.com/stm32-rs/stm32f3xx-hal/pull/47/files#diff-a4bc73d1077eb9a314216894b4643500R71)
However I can't justify why I set the default to 19. I think I saw it some other place and just implemented it here for good measure.

## Assuming your intentions
I assume you'd like to see something like:
```
#[repr(u8)]
pub enum SampleTime {
    T_1 = 0b000,
    T_2 = 0b001,
    // ...
    T_601 = 0b111,
``` 
Instead of implementing a `bitcode` function that returns these values.
The reason I decided for the latter was that 
a) we're already using the `bitcode` function in other scenarios here so it's more consistent with the other code
b) I wanted to avoid confusion regarding the (potentially future) calculation the actual sample time.

src/adc.rs
strom-und-spiele: Good catch. I didn't notice it when testing on hardware as my brain is not fast enough to actually see a difference ;)

strom-und-spiele:
> The current implementation does not consider using the other ADC prescaler[…] This is a bit more complicated to implement. 

Yes, I skipped this deliberately as I came to the same conclusion as you did ;)

src/adc.rs
strom-und-spiele: I'd like to postpone this to a later PR when we are resolvint the TODO of testing if rcc_hclk is equal sys_d1cpre_ck.

Sh3Rm4n: Fair enough

Sh3Rm4n: Can you explain, what you mean by this, exactly? What is rcc_hclk and what is sy_d1cpre_ck. 

I assume you mean, that rcc_hclk is the clock given by the AHB prescaler and what is sys_d1cpre_ck then? 

Sorry if this question is naiv, but i've difficulties by the naming to find these in the reference manual :)

src/adc.rs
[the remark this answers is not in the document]
Sh3Rm4n: Well #[repr(u8)] was the wrong blind guess then. 

I just wanted to say, that you use Prescale and SampleTime for two different value representations. Once to specify the values, which goes into the register, and then when using it to calculate timings. 

I just wanted to say, that the `From` trait would be the perfect candidate to convert between those two representations.

src/adc.rs
strom-und-spiele: TBH I just copy pasted from the [src that you linked to]:(https://docs.rs/stm32f3/0.10.0/src/stm32f3/stm32f303/adc1_2/ccr.rs.html#331-332)
```
#[doc = "1: Use AHB clock rcc_hclk3. In this case rcc_hclk must equal sys_d1cpre_ck"]
    SYNCDIV1 = 1,
```
As I'm not that good at remembering stuff and wanted to answer quickly I felt like this would be a reliable source for naming conventions. 

Now that I feel guilty for confuing you, lets specify: In [RM0316](https://www.st.com/resource/en/reference_manual/dm00043574.pdf) paragraph 15.3.3. reads:

> Software can use […] CKMODE[1:0]=01 only if the AHB prescaler of the 
RCC is set to 1 (the duty cycle of the AHB clock must be 50% in this configuration).

So that's all I meant by those confusing words above - thanks for tackling me from behind ;)

src/adc.rs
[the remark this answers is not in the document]
strom-und-spiele: I still don't get it. 
Let's collect what I think we agree on, so you can tell me what I misunderstood.
1. It makes sense to use one enum and one enum only for handling the prescaler. The same holds for the sample time.
2.  The two different value representations for the prescaler are necessary.
3. It's more intuitve to have the actual value of the prescaler as a plain integer than the bitcode. (Hence the plain definition in the enum definition and the `bitcode` function).
4. Allowing an enum to be casted to two different integer types while returing different values is a horrible idea and we should never ever do this.
5. It's more ideomatic (and less verbose) to write
```
enum A {
  first = 1,
  second = 2,
}
let new = A::first as u32;
```
than
```
enum A {
  first,
  second,
}
impl From<A> for u32 {
    fn from(a: A) -> u32 {
        match a {
          A::first => 1,
          A::second => 2,
        }
    }
}
let new = <u32>::from(A::first);
```
6. It's good practice to keep consistency. In this case: Also have a `bitcode` funciton for the sample time.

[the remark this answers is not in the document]
Sh3Rm4n: Thank you for the detailed answer. You've convinced me. I think my method does not really contribute to the readability of the code. So let's keep it the way with `bitcode`. 

I hope the discussion was not too exhausting. I have a tendency to make things overcomplicated just for being more "idomatic" (as of the perspective of my experience level).

Thank you for this great work :)

src/adc.rs
Sh3Rm4n: Ah okay. Thank you for the explanation. So this means, that HCLK_1 can only be used, if the AHB prescaler is set to 1. 

So `rcc.cfgr2.read().hpre().is_div1() == true` right?

src/adc.rs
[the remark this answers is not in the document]
strom-und-spiele: To be clear here: My answer was never about convincing anyone. I really appreciate the exchange of view points and you taking the time to check the code. No one here can have 5+ years of experience in Rust so let's try to learn and grow together.
On the other hand I don't think it makes sense to drag things out in a PR.

I also appreciated (and bookmarked) [your digression about debugging](https://github.com/stm32-rs/stm32f3xx-hal/pull/35), but feel that a PR is a bad place to put it (it might also be me having little experience with github)

Maybe @dfrankland agrees to activate the wiki so we can have discussions and agreements about idiomatic stm32f3xxx-code, debugging techniques and the like there.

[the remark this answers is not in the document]
Sh3Rm4n: This would be an interesting topic to discuss in general across all stm32xxxxx-hal implementations. (so many x's 😅)

I'm in favor of establishing some kind of alignment across these implementations. 

But as of now, I don't know where to start, or if this alignment is already happening. I'll have to ask in the next embedded meeting. I'm sure they already thought about this. 

At least some kind of guideline for a hal implementation would be great.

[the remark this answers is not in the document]
strom-und-spiele: I totally agree. Having a guideline makes decisions so much easier :D

Are those "embedded meetings" local or virtual? I'd really like to attend (maybe after the toilet paper crisis is over).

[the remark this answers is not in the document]
Sh3Rm4n: The [rust-embedded wg](https://github.com/rust-embedded/wg) does have a [matrix channel](#rust-embedded:matrix.org).  They do have a meeting every Tuesday 8pm CET over the matrix channel.

As there are many people, which are also involved into the stm32-rs project, this seems to be the place, to at least ask and contact members of the stm32-rs project.

strom-und-spiele:
> Could you remove [36e02c6](https://github.com/stm32-rs/stm32f3xx-hal/commit/36e02c69714f15c37d2ffdbfd1c46f610a73e3ef) and [7c7f8b4](https://github.com/stm32-rs/stm32f3xx-hal/commit/7c7f8b48746e785200cd1c461cd5735197d6a1dc). I assume these are from #38, so i like to keep them there :)

I have no clue how to do this without doing nasty stuff. 
As those changes are already merged into master is it ok to just ignore this fauxpas?

Sh3Rm4n:
Have a look at [git rebase](https://git-scm.com/book/en/v2/Git-Branching-Rebasing). Pretty useful subcommand. Though the interesting part enfoldes, when you use `git rebase --interactive <commit hash you want to rebase to. 

There you have the option to remove commits, change the message of the commits, reorder commits and more. Here is another official article about [rewriting the git history](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History). 

Remeber, you have to force push to the branch, when changing the commit history. I suggest using [`--force-with-lease`](https://blog.developer.atlassian.com/force-with-lease/) as the safer alternative to `--force`

Hope this helps :) 

> As those changes are already merged into master is it ok to just ignore this fauxpas?

I assume that they would be ignored, but they still show up in the "files changed" part, so I'm not really sure about this.

strom-und-spiele:
Huge thanks to @Sh3Rm4n for reviewing and teaching me about rebasing.

src/lib.rs
```diff
@@ -3,17 +3,10 @@
...
-        stm32f373
-        stm32f378
-        stm32f334
-        stm32f328
-        stm32f358
-        stm32f398"
+    "This crate requires you to specify your target device as a feautre.  
+    Note that the more specific you pick the device, the more functionality will be available.
+
+    See `Cargo.toml` section 'features' for a comprehensive list of devices (starting with stm32f...)."
```
Sh3Rm4n, line 9: Hm. Not sure if it should stay this way. The list from before is outdated so this change overall is a good idea. But I would'nt point a user from  this create to look at the Cargo.toml  / source code to find out the relevant information. I would rather have the information available through the README or in some kind of way, so the user could find it at docs.rs. 

I know you tackle this in https://github.com/stm32-rs/stm32f3xx-hal/pull/75, so deleting this line for now should be okay. If could stay as well for now, and we could change this in the mentioned PR :)

src/adc.rs
Sh3Rm4n: Is this comment addressed?

src/lib.rs
strom-und-spiele: removed it and added a link to #75 in the commit message

strom-und-spiele:
@https://github.com/stm32-rs/stm32f3xx-hal/pull/47#discussion_r404340511 Thanks for catching that typo (again) is fixed in https://github.com/stm32-rs/stm32f3xx-hal/pull/47/commits/bedfa0f6a107a0cb2a03e6ff2c4d611cbff8b6d4
the code does not fit the issue though - did you mean something else?

src/adc.rs
strom-und-spiele: hmm it got lost in the rebase action - anyhow I like the new solution (https://github.com/stm32-rs/stm32f3xx-hal/pull/47/commits/efe8f23ced863e97c18b670a545722d6b82f5557) more

src/adc.rs
strom-und-spiele: I [implemented the use of `CKMODE`](https://github.com/stm32-rs/stm32f3xx-hal/pull/47/commits/efe8f23ced863e97c18b670a545722d6b82f5557) and I like it more this way. But maybe there is an even more elegant way to go about it?

Using `rcc` is not possible, as it is forzen in `clocks`, so I implemented it differently.
I also added a check to verify the user is not trying to set up ADC1 and ADC2 with different settings (as this is not possible)

Sh3Rm4n:
Hey I tested your example yesterday and it worked for me on my f3 discovery board! :) 

 I did write down some notes for review but this is still in progress. 

However two things that came to my mind:

Can you be more explicit, that the pa0 you are using in the example is the user button on the discovery board. It is pretty clever to use it, btw. That means that every one who owns a discovery board can test the example immediately. 

And the user should now, that the ADC has a 12 bit resolution, so that the maximum expected value is 4095 :)

strom-und-spiele:
> Can you be more explicit, that the pa0 you are using in the example is the user button on the discovery board. It is pretty clever to use it, btw. That means that every one who owns a discovery board can test the example immediately.
> 
> And the user should now, that the ADC has a 12 bit resolution, so that the maximum expected value is 4095 :)
Thanks for the input - it made me think how much information to include in the output and in the comments. where to spread it and then I went down the rabbit hole of explaining how to properly convert adc values.
Not sure I like it that way but now we can talk about something…

ra-kete:
@strom-und-spiele What's the status on this PR? I haven't look at the code again, but from the comments here it looks like it's almost ready to merge. Would be great if we could get this over the finish line, ADC support is something I'd personally like to see too.

strom-und-spiele:
I've been using it productively as a base for my ADC needs ever since. 
I'll set aside some time on the weekend to rebase it and we can merge it (for stm32f303 only btw)

ra-kete:
Awesome! I'll give the code a quick review today then.

src/adc.rs
```diff
@@ -0,0 +1,562 @@
...
+                    // bit is cleared by hardware
+                    this_adc.wait_adc_clk_cycles(4);
+                    this_adc.enable();
+                    Some(this_adc)
+                }
+
+                /// Software can use CKMODE::SYNCDIV1 only if
+                /// hclk and sysclk are the same. (see reference manual 15.3.3)
+                fn clocks_welldefined(&self, clocks: Clocks) -> bool {
+                    !(self.ckmode == CKMODE::SYNCDIV1 && !(clocks.hclk().0 == clocks.sysclk().0))
```
ra-kete, line 338: I really had trouble understanding this condition. I think this would really benefit from being a bit more verbose, like:
```rust
if self.ckmode == CHKMODE::SYNCDIV1 {
    clocks.hclk().0 == clocks.sysclk().0
} else {
    true
}
```
Or, if it absolutely needs to be one line, at least get rid of the double negation:
```rust
self.ckmode != CKMODE::SYNCDIV1 || clocks.hclk().0 == clocks.sysclk().0
```

src/adc.rs
```diff
@@ -0,0 +1,562 @@
...
+                }
+
+                /// sets up adc in one shot mode for a single channel
+                pub fn setup_oneshot(&mut self) {
+                    // stop and clear overrun events
+                    self.rb.cr.modify(|_, w| w.adstp().set_bit());
+                    self.rb.isr.modify(|_, w| w.ovr().clear_bit());
```
ra-kete, line 345: I think clearing the bit is wrong here. According to the RM: "OVR: ADC overrun. This bit is set by hardware when an overrun occurs on a regular channel, meaning that a new conversion has completed while the EOC flag was already set. **It is cleared by software writing 1 to it.**" So, to clear the overrun flag, you have to actually set this bit.

If registers have more descriptive methods than the generic `set_bit` and `clear_bit`, it is usually a good idea to use these. It prevents bugs like this one and also improve readability. You use `set_bit` and `clear_bit` pretty consistently here. I'd suggest converting them all to the descriptive methods, if possible.

src/adc.rs
```diff
@@ -0,0 +1,562 @@
...
+/// TODO: Implement other modes (DMA, Differential,…)
+pub enum OperationMode {
+    OneShot,
+}
+
+#[derive(Clone, Copy, PartialEq)]
+/// ADC CKMODE
+///
+/// TODO: Add ASYNCHRONOUS mode
+pub enum CKMODE {
```
ra-kete, line 107: Rust coding conventions suggest using CamelCase for type names. So something like `Ckmode` or `CkMode` would fit better

src/adc.rs
```diff
@@ -0,0 +1,562 @@
...
+
+impl Default for Align {
+    fn default() -> Self {
+        Align::Right
+    }
+}
+
+impl Align {
+    /// Conversion to bits for ALIGN in ADCx_CFGR
+    fn bitvalue(&self) -> bool {
```
ra-kete, line 149: Consider replacing this with an `impl From<Align> for ALIGN_A`, like you did with CKMODE.

src/adc.rs
```diff
@@ -0,0 +1,562 @@
...
+                fn enable(&mut self) {
+                    self.rb.cr.modify(|_, w| w.aden().set_bit());
+                    while self.rb.isr.read().adrdy().bit_is_clear() {}
+                }
+
+                fn disable(&mut self) {
+                    self.rb.cr.modify(|_, w| w.aden().clear_bit());
+                }
+
+
```
ra-kete, line 370: Superfluous newline. Pretty sure the only reason rustfmt is okay with this is because it doesn't handle macros

src/adc.rs
```diff
@@ -0,0 +1,562 @@
...
+                fn wait_adc_clk_cycles(&self, cycles: u32) {
+                    let adc_clk_cycle = self.clocks.hclk().0 / (self.ckmode as u32);
+                    asm::delay(adc_clk_cycle * cycles);
+                }
+
+                fn advregen_enable(&mut self){
+                    // need to go through intermediate first
+                    self.rb.cr.modify(|_, w| w.advregen().intermediate());
+                    self.rb.cr.modify(|_, w| w.advregen().enabled());
+
```
ra-kete, line 398: Superfluous newline

src/adc.rs
```diff
@@ -0,0 +1,562 @@
...
+                    self.rb.cr.modify(|_, w| w
+                        // NOTE: needs to be adopted if implementing differential input
+                        .adcaldif().clear_bit()
+                        .adcal()   .set_bit());
+
+                    while self.rb.cr.read().adcal().bit_is_set() {}
+                }
+
+
```
ra-kete, line 388: Superfluous newline

src/adc.rs
```diff
@@ -0,0 +1,562 @@
...
+                    self.rb.cr.modify(|_, w| w.advregen().intermediate());
+                    self.rb.cr.modify(|_, w| w.advregen().enabled());
+
+                }
+
+                /// wait for the advregen to startup.
+                ///
+                /// This is based on the MAX_ADVREGEN_STARTUP_US of the device.
+                fn wait_advregen_startup(&self) {
+                        asm::delay((MAX_ADVREGEN_STARTUP_US * 1_000_000) / self.clocks.sysclk().0);
```
ra-kete, line 405: The identation is wrong here

src/adc.rs
```diff
@@ -0,0 +1,562 @@
...
+
+                /// busy ADC read
+                fn convert_one(&mut self, chan: u8) -> u16 {
+                    self.ensure_oneshot();
+                    self.set_chan_smps(chan, SampleTime::default());
+                    self.select_single_chan(chan);
+
+                    self.rb.cr.modify(|_, w| w.adstart().set_bit());
+                    while self.rb.isr.read().eos().is_not_complete()
+                        {}
```
ra-kete, line 416: Man, I really wish rustfmt would work inside macros. Feeling like a jackass nitpicking about all these format issues... But those braces should go on the previous line

src/adc.rs
```diff
@@ -0,0 +1,562 @@
...
+                fn ensure_oneshot(&mut self) {
+                    match self.operation_mode {
+                        Some(mode) =>
+                        {
+                            if mode != OperationMode::OneShot {
+                                self.setup_oneshot();
+                            }
+                        },
+                        _ => self.setup_oneshot(),
+                    };
```
ra-kete, line 430: This can be simplified:
```
if self.operation_mode != Some(OperationMode::OneShot) {
    self.setup_oneshot();
}
```

src/adc.rs
```diff
@@ -0,0 +1,562 @@
...
+///
+/// TODO: Extend/differentiate beyond f303.
+macro_rules! adc12_hal {
+    ($(
+            $ADC:ident: ($adcx:ident),
+    )+) => {
+        $(
+            impl Adc<$ADC> {
+                fn enable_clock(&self, ahb: &mut AHB, adc_common: &mut ADC1_2) -> bool {
+                    if ahb.enr().read().adc34en().is_enabled() &&
```
ra-kete, line 494: This should be `adc12en()`, right?

src/adc.rs
```diff
@@ -0,0 +1,562 @@
...
+                fn enable_clock(&self, ahb: &mut AHB, adc_common: &mut ADC3_4) -> bool {
+                    if ahb.enr().read().adc34en().is_enabled() {
+                        return (adc_common.ccr.read().ckmode().variant() == self.ckmode.into())
+                    }
+                    ahb.enr().modify(|_, w| w.adc34en().enabled());
+                    adc_common.ccr.modify(|_, w| w
+                        .ckmode().variant(self.ckmode.into())
+                    );
+                    true
+                }
```
ra-kete, line 539: For consistency it would be nice if this function and the equivalent in `adc12_hal!` would have the same formatting.

src/adc.rs
```diff
@@ -0,0 +1,562 @@
...
+    feature = "stm32f303xe",
+))]
+macro_rules! adc34_hal {
+    ($(
+            $ADC:ident: ($adcx:ident),
+    )+) => {
+        $(
+            impl Adc<$ADC> {
+                /// Returns true iff
+                ///     the clock was enabled
```
ra-kete, line 528: Should this say "the clock was *disabled*"?

src/adc.rs
```diff
@@ -0,0 +1,562 @@
...
+//! Note that the more specific your hardware selection is
+//! (e.g. stm32f303**xc** instead of just stm32f303)
+//! the more functionality is accessible
+//! (in this case: ADC3 and ADC4 additionaly to ADC1 and ADC2).
+//!
+//! # Examples
+//! For a simple, working example check `adc.rs` in the examples folder.
+use cortex_m::asm;
+use embedded_hal::adc::{Channel, OneShot};
+#[cfg(feature = "stm32f303")]
```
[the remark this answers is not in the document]
ra-kete: Okay, not sure what I was seeing, the module is actually gated on "device-selected" now. Still, I think it would make sense to gate it on "stm32f303" and remove those gates here.

src/adc.rs
```diff
@@ -0,0 +1,562 @@
...
+    T_7,
+    T_19,
+    T_61,
+    T_181,
+    T_601,
+}
+
+impl Default for SampleTime {
+    fn default() -> Self {
+        SampleTime::T_19
```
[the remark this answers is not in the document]
strom-und-spiele: The RM stated:
> For each channel, SMP[2:0] bits must be programmed to respect a minimum sampling time as specified in the ADC charateristics section of the datasheets.

So to avoid any problems there and to have a stable measurement, I decided to set it to T_19. Weather this is a sensible choice is up to debate, but I don't agree that using T_1 is a better one, as this is a *default* and not a *reset* value we're setting.

src/adc.rs
```diff
@@ -0,0 +1,562 @@
...
+    gpioa::PA5<Analog> => 2_u8,
+    gpioa::PA6<Analog> => 3_u8,
+    gpioa::PA7<Analog> => 4_u8,
+    gpioc::PC4<Analog> => 5_u8,
+    gpioc::PC5<Analog> => 11_u8,
+    gpiob::PB2<Analog> => 12_u8,
+    gpioc::PC0<Analog> => 6_u8,
+    gpioc::PC1<Analog> => 7_u8,
+    gpioc::PC2<Analog> => 8_u8,
+    gpioc::PC3<Analog> => 9_u8,
```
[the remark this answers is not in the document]
strom-und-spiele: I sorted ADC1 by channel number and later ADC12 (shared channels) by channel number.
However you're right that it's easier to read and find missing ones.

src/adc.rs
strom-und-spiele: Thanks for spotting this. When I first wrote this module, the old PAC didn't provide the more descriptive methods. 
I'm all in with using them.

src/adc.rs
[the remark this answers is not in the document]
ra-kete: I agree that we should deviate from the device's defaults if we can find safer ones. However, in this case we should:

a) Add a comment explaining why we choose this default
b) Make sure that T_19 is actually sensible

For b), the datasheet of the discovery says minimum sampling time is `1.5 / (ADC clock frequency)`. Which means it needs at minimum 1.5 ADC clock cycles for each sample, right? Which means at least for the discovery `T_1` would actually be the sensible default.

Now, to do this right we'd actually have to go through all stm32f303 datasheets and check their ADC characteristics. If we don't want to do that, we should either use the device default (`T_1`), or use the highest value just to be sure (`T_601`). Personally I think using the device default will usually be fine. I would rather not slow the sampling down to it's maximum if it's probably not necessary. But I also don't like randomly picking a value from the middle without any rationale.

src/adc.rs
```diff
@@ -374,11 +378,10 @@ macro_rules! adc_hal {
...
 
                     self.rb.cr.modify(|_, w| w
-                        // NOTE: needs to be adopted if implementing differential input
-                        .adcaldif().clear_bit()
-                        .adcal()   .set_bit());
+                        .adcaldif().single_ended()
+                        .adcal()   .calibration());
 
-                    while self.rb.cr.read().adcal().bit_is_set() {}
+                    while !self.rb.cr.read().adcal().is_complete() {}
```
ra-kete, line 381: There is no `is_not_complete`, because it's called `is_calibration` ;)

src/adc.rs
```diff
@@ -0,0 +1,549 @@
...
+                    self.rb.cfgr.modify(|_, w| w.align().variant(align.into()));
+                }
+
+                fn enable(&mut self) {
+                    self.rb.cr.modify(|_, w| w.aden().enable());
+                    while self.rb.isr.read().adrdy().is_not_ready() {}
+                }
+
+                fn disable(&mut self) {
+                    self.rb.cr.modify(|_, w| w.aden().clear_bit());
```
ra-kete, line 365: I haven't actually checked this, but I'm pretty sure there is no `disable` method, because you cannot disable the ADC via the `ADEN` bit. You need to use `ADDIS`, which also has a `disable` method.

src/adc.rs
```diff
@@ -336,27 +336,31 @@ macro_rules! adc_hal {
...
                     );
 
-                    self.rb.sqr1.modify(|_, w| w.l().bits(0));
+                    self.set_sequence_len(1);
 
                     self.operation_mode = Some(OperationMode::OneShot);
                 }
 
+                fn set_sequence_len(&mut self, len: u8) {
+                    debug_assert!(len <= 16);
```
ra-kete, line 351: Since you use `len - 1` below, you need to check that `len - 1 < 16`. Or add an additional check that `len > 0`.

src/adc.rs
```diff
@@ -0,0 +1,549 @@
...
+                    this_adc.enable();
+
+                    this_adc
+                }
+
+                /// Software can use CkMode::SYNCDIV1 only if
+                /// hclk and sysclk are the same. (see reference manual 15.3.3)
+                fn clocks_welldefined(&self, clocks: Clocks) -> bool {
+                    if (self.ckmode == CkMode::SYNCDIV1)
+                    {
```
ra-kete, line 328: Formatting: That opening brace should be on the previous line

PR #50, opened 2020-01-05
Enable blocking.Write for UART, replace Void with Infallible 

joestarwalker:


dfrankland:
This looks good. I'll need to run a quick regression before merging.

dfrankland:
@joestarwalker one quick change request: can you add to the CHANGELOG.md?

dfrankland:
Thank you @joestarwalker!

PR #60, opened 2020-02-16
Fix SPI initialization to be MSB first

### LINKED ISSUE #59: SPI initialization should set CR1.LSBFIRST bit to 'MSB first' by default
Many SPI protocol use 'MSB first' in MOSI/MISO line when transfer bits, so 'MSB first' should be set in the SPI_CR1.LSBFIRST bit by default.

I see https://github.com/stm32-rs/stm32f3xx-hal/blob/master/src/spi.rs#L152 try to do the right thing:
````rust
// LSBFIRST: MSB first
````
but https://github.com/stm32-rs/stm32f3xx-hal/blob/master/src/spi.rs#L185 acturally set is to 'LSB first'
```rust
                            .lsbfirst()    // line 184, return stm32f3::stm32f303::spi1::cr1::LSBFIRST_W<'_>
                            .lsbfirst()    // line 185, call method on struct returned from line 184
```

So line 185 should be changed to `.msbfirst()`.

I have successful communicated to my SSD1306 OLED display module from a STM32F3Discovery board after making the this change.

_Note: for the SPI to work, the master branch should be used, as another SPI initialization bug fixed in #35 haven't landed on '0.4.0' release yet._

arlosi:
The comments indicate that the code is attempting to set the SPI order to MSB first, however it actually sets it to LSB first. Since most devices are MSB first, this is the correct choice.

Closes #59.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,6 @@ rust:
   - stable
   - nightly
 
-cache: cargo
-
 matrix:
   allow_failures:
     - rust: nightly
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Implement blocking Write for UART
 - Implement blocking Read for I2C
 
+### Fixed
+- Regression in v0.4.0 that set SPI to LSB-first ordering
+
 ## [v0.4.0] - 2019-12-27
 
 ### Added
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -182,7 +182,7 @@ macro_rules! hal {
                         w.spe()
                             .enabled()
                             .lsbfirst()
-                            .lsbfirst()
+                            .msbfirst()
                             .ssi()
                             .slave_not_selected()
                             .ssm()


arlosi:
With master + this change, I can successfully communicate with two different SPI devices using an STM32F334

Sh3Rm4n:
Thank you for this PR. Sorry for the delay, I was quite busy with work and uni stuff the last weeks. 

It seems like I oversaw this, when I changed this line of code. I briefly tested it today, and will look into this PR this week :)

Sh3Rm4n:
Okay I've confirmed that your fix is working. 

Would be strange, if not :sweat_smile: 

I would be happy, if you added a note to the changelog. Something like:

```
# Fix

- Fixed regression ... spi was not correctly intialized as master
```

:)

Sh3Rm4n:
Thanks!

Meh travis timed out. Force pushing with `--amend --no-edit` or squashing the commits should retrigger travis. 

I should look into fixing travis sometime, as two instances doing the sames is really not useful :sweat_smile:

Sh3Rm4n:
Thanks for force pushing. This sadly didn't help. The travis instance has a hard time initializing sscache. I'll look into fixing this.

Sh3Rm4n:
Disable caching in travis worked :tada:

Issue #57, opened 2020-02-07
stm32f3xx_hal::rcc::CFGR::freeze() gives wrong default clocks: it gives half of the real clocks


zhenkyle:
Today when I'm doing a serial project with the `usart1` hardware on my f3discovery board, to get the default clock configuration by the following code:
```
let clocks = rcc.cfgr.freeze(&mut flash.acr);
```
I can't get anything over serial line, after I debuged the hardward by a logic analyzer, I found out the actural bps of the `usart1` hardware is twice as much as the intended bps value.

gdb give this clue, showing the hardware is runing on 4MHz sysclk:
```
(gdb) print clocks
$1 = stm32f3xx_hal::rcc::Clocks {hclk: stm32f3xx_hal::time::Hertz (4000000), pclk1: stm32f3xx_hal::time::Hertz (4000000), pclk2: stm32f3xx_hal::time::Hertz (4000000), ppre1: 1, ppre2: 1, sysclk: stm32f3xx_hal::time::Hertz (4000000), usbclk_valid: false}
```
which is curious, because I was told the default sysclk is 8MHz. If the default clock configuration gives half of the real clocks, it can explain the double of bps in `usart1`.

Freeze the clock configuration with 8MHz sysclk manually:
```
let clocks = rcc.cfgr.sysclk(8.mhz()).freeze(&mut flash.acr);
```
everythings works fine.

Switching to `stm32f30x-hal` crate and freeze with default clock configuration again, everything works fine too. And gdb gives me 8MHz sysclk.

The `freeze()` code in `rcc.rs` is quite complex, I haven't figure it out yet. but I think there's some thing wrong with `PLLMUL` setting, in the reference manual, the PLL multiplication factor can never be 1x.

ra-kete:
AFAICT the issue is that (at least on some devices*)the HSI clock is divided by 2 when used as PLLSRC, but when the HSI is used as the system clock directly (i.e. PLL is not enabled), it is not devided.

In rcc.rs, the code that decides if the PLL is used and what the final SYSCLK will be is in `calc_sysclk` and `calc_pll`. Here is a simplified version of that logic (ignoring the HSE and PLL options selection):

```rust
const HSI: u32 = 8_000_000;

fn calc_sysclk(&self) -> (u32, Option<...>) {
    let pllsrcclk = HSI / 2;
    let pllmul = self.sysclk.unwrap_or(pllsrcclk) / pllsrcclk;

[...]

Issue #62, opened 2020-03-05
Module gpioe is not found


smedellin90:
```
#![no_main]
#![no_std]

use embedded_hal::digital::v1_compat::{OldInputPin, OldOutputPin};
use rtfm::app;
use portable::button;
use stm32f3xx_hal::prelude::*;
use stm32f3xx_hal::{delay, gpio, i2c ,spi, stm32, timer};

#[cfg(not(test))]
extern crate panic_semihosting;

type Button0Pin = gpio::gpioa::PA6<gpio::Input<gpio::PullUp>>;
type BlueLED = gpio::gpioe::PE8<gpio::Output<gpio::PushPull>>;
```

There is an error in type BlueLED. The module **gpioe** cannot be found within the **gpio** module. May someone please advise?

Sh3Rm4n:
Which microcontroller are you using / what feature did you enable in your Cargo.toml?

There is quite a bit of macro work involved creating the gpio modules depending on the feature. 

https://github.com/stm32-rs/stm32f3xx-hal/blob/609af0a91c2030ff2c3e35705e5394d66ec3cb81/src/gpio.rs#L1732

Here is a list of available features. The feature you choose has to match the microcontroller, for which you are planning to build it. 
https://github.com/stm32-rs/stm32f3xx-hal/blob/609af0a91c2030ff2c3e35705e5394d66ec3cb81/Cargo.toml#L51

smedellin90:
@Sh3Rm4n , this is my Cargo.toml:
`[package]
name = "embedded_testbench"
version = "0.1.0"
authors = ["Scott"]
edition = "2018"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[workspace]

[dependencies]
stm32f3xx-hal = { version ="0.4.0", features = ["rt", "stm32f303"] }
##stm32f30x-hal = { version = "0.2", features = ["rt", "stm32f303" ] }
# = { version = "0.5", features = ["rt", "stm32f103" ] }
cortex-m = "0.6"
#cortex-m-rt = "0.6.3"
##cortex-m-rt = { version = "0.6", features = ["device"] }
cortex-m-rtfm = "0.5.0"
panic-semihosting = "0.5"
#bme280 = "0.2"
#pwm-speaker = { git = "https://github.com/TeXitoi/pwm-speaker", branch = "master" }
[...]

PR #67, opened 2020-03-08
Rework PLL calculation and utilize oscillator clock divison

Sh3Rm4n:
# Rework

This PR turned out to be rework of the PLL setup.

Now [`get _sysclk`](https://github.com/stm32-rs/stm32f3xx-hal/pull/67/files#diff-c6941ad2425cd611249033bed1925f02R411) will choose, whether PLL is needed as the source for the system clock.

The [`usb_valid`](https://github.com/stm32-rs/stm32f3xx-hal/pull/67/files#diff-c6941ad2425cd611249033bed1925f02R521) check should work just fine, as we need the PLL anyways to get the supported frequencies (48 MHz and 72 MHz). Both can not be  directly generated from the external oscillator as the `HSE OSC` does only allow up to 32 MHz. 

`calc_pll` does now calculate the optimal divisor and multiplier values to generate the system clock, which the user provides. For that the [greatest common divisor](https://en.wikipedia.org/wiki/Greatest_common_divisor) calculation is used, because the range of the possible values, which the PRE_DIV and PLL_MUL can take is 16 in both cases.

`calc_pll` is split into 2 implementations, [one](https://github.com/stm32-rs/stm32f3xx-hal/pull/67/files#diff-c6941ad2425cd611249033bed1925f02R277) for [devices](https://www.st.com/content/ccc/resource/technical/document/reference_manual/4a/19/6e/18/9d/92/43/32/DM00043574.pdf/files/DM00043574.pdf/jcr:content/translations/en.DM00043574.pdf#page=128
) which only can adjust the division of the clock of the external oscillator.
The [other one](https://github.com/stm32-rs/stm32f3xx-hal/pull/67/files#diff-c6941ad2425cd611249033bed1925f02R354), where both [HSI and HSE](https://www.st.com/content/ccc/resource/technical/document/reference_manual/4a/19/6e/18/9d/92/43/32/DM00043574.pdf/files/DM00043574.pdf/jcr:content/translations/en.DM00043574.pdf#page=127) division can be adjusted.

> Both can not be  directly generated from the external oscillator as the `HSE OSC` does only allow up to 32 MHz. 

Now that the divisor (PRE_DIV) is utilized, 48 MHz and 72 MHz can be generated out of a 32 MHz

`48 MHz / 32 MHz` can be simplified to `3 / 2` as the greatest common divisor is `16`. `PLL_MUL` is now set to `3` and PRE_DIV is set to `2`. 

Similarly, `72 MHz / 32 MHz` can be simplified to `9 / 4` as the greatest common divisor is `8`.

# Fix

Initially a fix, which removes HSI division for devices, which do not divide HSI. 

Fix https://github.com/stm32-rs/stm32f3xx-hal/issues/57

Ref https://github.com/stm32-rs/stm32f3xx-hal/issues/57#issuecomment-590063873

Relevant parts in the reference manual: 

[with `/2` divisor after HSI](https://www.st.com/content/ccc/resource/technical/document/reference_manual/4a/19/6e/18/9d/92/43/32/DM00043574.pdf/files/DM00043574.pdf/jcr:content/translations/en.DM00043574.pdf#page=126)

[without `/2` divisor after HSI](https://www.st.com/content/ccc/resource/technical/document/reference_manual/4a/19/6e/18/9d/92/43/32/DM00043574.pdf/files/DM00043574.pdf/jcr:content/translations/en.DM00043574.pdf#page=127)


Sh3Rm4n:
Reminder: Compare with `stm32f30x-hal`, as this crate does not have this bug.

Sh3Rm4n:
For stm32f302 devices there is also a distinction between [B/C](https://www.st.com/content/ccc/resource/technical/document/reference_manual/16/bf/04/cd/f6/25/44/a7/DM00094349.pdf/files/DM00094349.pdf/jcr:content/translations/en.DM00094349.pdf#page=116) and [D/E](https://www.st.com/content/ccc/resource/technical/document/reference_manual/16/bf/04/cd/f6/25/44/a7/DM00094349.pdf/files/DM00094349.pdf/jcr:content/translations/en.DM00094349.pdf#page=117) devices. 

This distinction between these devices do not exist yet.

Sh3Rm4n:
Going through these quite complicated clock trees: Another reminder. 

[stm32f303d/e](https://www.st.com/content/ccc/resource/technical/document/reference_manual/4a/19/6e/18/9d/92/43/32/DM00043574.pdf/files/DM00043574.pdf/jcr:content/translations/en.DM00043574.pdf#page=127) does have an extra PREDIV block, while [stm32f302d/e](https://www.st.com/content/ccc/resource/technical/document/reference_manual/16/bf/04/cd/f6/25/44/a7/DM00094349.pdf/files/DM00094349.pdf/jcr:content/translations/en.DM00094349.pdf#page=118) does not. 

stm32f303d/e:

img-2020-03-12-235719

stm32f302d/e:

img-2020-03-12-235731

Cargo.toml
```diff
@@ -52,6 +52,10 @@ rt = ["stm32f3/rt"]
 stm32f301 = ["stm32f3/stm32f301", "device-selected"]
 stm32f318 = ["stm32f3/stm32f3x8", "device-selected"]
 stm32f302 = ["stm32f3/stm32f302", "device-selected"]
```
Sh3Rm4n: With the changes in clock calculation, this would implicitly "choose" stm32f302xb / stm32f302xc for the clock calculation. Is this desired behavior? Do we want to force the user to specify the device if needed (i assume this will also happen for other devices in the future) or do we implicitly choose one? This could lead to unexpected results for the user as some registers could be set to wrong states.

strom-und-spiele: I'd vouch for forcing the user to specify the device (i.e. removing "device-selected" from stm32f303 and adding it to all the sub-variants). 
I expect 
* Less people getting confused (see e.g. https://github.com/stm32-rs/stm32f3xx-hal/issues/62)
* Some backwards comparability getting broken (but that's fine)
* More awareness when implementing stuff
* More stuff getting implemented (as it is easier to adopt something for a sub-variant if there is already the infrastructure for selecting only this variant)

It'd like to see
* a throughout instructions in the README on how to select the proper decive
* a short note to read the mentioned README instructions when [no sub-variant is used](https://github.com/stm32-rs/stm32f3xx-hal/blob/14ed7e55ed8f83f09a54d08fc81d7575d6404403/src/lib.rs#L5) 
(side note: I proposed something along those lines in https://github.com/stm32-rs/stm32f3xx-hal/pull/47 - see [the errormsg there](https://github.com/strom-und-spiele/stm32f3xx-hal/blob/fb5d6a7a1aa5d6b31eedbc7ac434f5e4bf1010a5/src/lib.rs#L5) )

Sh3Rm4n: Good point. Confusing users when falling back implicitly to a specific device is definitely not the way to go. We should handle every sub-variant as and independent device. This would make some parts of this HAL more verbose, but then it's easier to reason about little differences between those devices.

I'd do this with a cleanup of the README in a follow up PR, as it would not hurt, that the README can get a bit more information.

strom-und-spiele: Actually we don't need to be more verbose. See my proposal in https://github.com/stm32-rs/stm32f3xx-hal/pull/75

Sh3Rm4n:
So I tested it on different clock configuration on hardware, seeing if the clock configuration is correctly initialized and most things worked as expected. Now it is possible to choose clocks like 14 MHz on an 8 MHz based Oscillator clocks. 

However, with 9 MHz I had a strange problem I used itm (`iprintln!`) to print out the clock configuration return from `freeze()`. I always had to adjust the gdb config according to the system clock.

```
monitor tpiu config internal /tmp/itm.fifo uart off 9000000
```

But on 9 MHz `itmdump` return garbage. I don't really know why, I haven't looked into much further. But looking into the register settings and variables via gdb, everything was set up correctly.

Sh3Rm4n:
I feel like this is ready to be review again. 

I don't like the way that the rcc module contains so many asserts, which keeping coming up, when using a wrong configuration. A better documentation for valid clock settings or a way to propagate the error to the user is needed. But this is not scope of this PR. 

@russellmcc maybe you'd like to take another look. If not I'll ask somebody else. I'm comfortable merging this code as is but would like to have it reviewed, as this is code everybody depends upon, when using this hal.

PR #68, opened 2020-03-08
Update changelog with MonoTimer frequency fix

ra-kete:
Following up on #56 to also update the changelog.


Sh3Rm4n:
Cool, thank you :)

PR #72, opened 2020-03-13
Fixes for i2c Read 

tstellanova:
- Use automatic mode for autoend with `Read`
- Check for variant success first in `busy_wait` macro

Tested with the following embedded hal i2c drivers:
- bmp280 (barometer)
- em7180 (imu/ahrs)
- bno080 (ahrs wip)

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -88,12 +88,12 @@ macro_rules! busy_wait {
         loop {
             let isr = $i2c.isr.read();
 
-            if isr.berr().is_error() {
+            if isr.$flag().$variant() {
+                break;
+            } else if isr.berr().is_error() {
                 return Err(Error::Bus);
             } else if isr.arlo().is_lost() {
                 return Err(Error::Arbitration);
-            } else if isr.$flag().$variant() {
-                break;
             } else {
                 // try again
             }
@@ -228,7 +228,7 @@ macro_rules! hal {
                             .start()
                             .start()
                             .autoend()
-                            .software()
+                            .automatic()
                     });
 
                     for byte in buffer {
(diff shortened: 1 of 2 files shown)


Sh3Rm4n:
Cool! :+1: Thanks for this PR. I'll have a closer look later today  (i hope)

tstellanova:
Thanks!  Here's an example analyzer trace from where `write_read` was failing: 
failed_write_read_bno080

tstellanova:
And an example trace from where read was failing due to an incorrect autoend setting: 
read_4a_ack_fail

src/i2c.rs
```diff
@@ -227,7 +227,7 @@ macro_rules! hal {
                             .start()
                             .start()
                             .autoend()
-                            .software()
+                            .automatic()
```
Sh3Rm4n: Good catch. Without it the slave on the bus would never know, when to stop writing to master.

src/i2c.rs
```diff
@@ -88,12 +88,12 @@ macro_rules! busy_wait {
         loop {
             let isr = $i2c.isr.read();
 
-            if isr.berr().is_error() {
```
Sh3Rm4n: I'm unsure about this change though. This change assumes that the variant flag is only set, when no error occured, otherwise errors would be ignored. 

I do not if this is the case, though. I'm currently reading into the documentation, but maybe you can point me into the right direction :)

Sh3Rm4n:
Thank you for your logic analyzer proves btw :) 

If you want, you could a a FIX note to the CHANGELOG :)

src/i2c.rs
tstellanova: My understanding of the purpose of this macro is to first check whether the variant criterion is satisfied, and secondarily, if the criterion is not satisfied, check for errors.  My change to the macro crystallizes these priorities.

tstellanova: I'm not too attached to this change-- just let me know if you want me to revert it.

Sh3Rm4n:
The CI is currently stuck. I don't know, how to rerun it, but as conflicts have to be resolved anyways, this should retrigger the CI. I hope it will work then :)

tstellanova:
Fixed the merge conflicts and re-squashed the PR

PR #73, opened 2020-03-15
Change Into trait to From trait

Sh3Rm4n:
Change the `Into` traits to `From` traits, as `Into` is automatically available if `From` is implemented. `Into` should only be used if  `From` can not be implemented due to the orphan rule.

### DIFF
--- a/src/time.rs
+++ b/src/time.rs
@@ -53,21 +53,21 @@ impl U32Ext for u32 {
     }
 }
 
-impl Into<Hertz> for KiloHertz {
-    fn into(self) -> Hertz {
-        Hertz(self.0 * 1_000)
+impl From<KiloHertz> for Hertz {
+    fn from(val: KiloHertz) -> Self {
+        Self(val.0 * 1_000)
     }
 }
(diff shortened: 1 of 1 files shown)


strom-und-spiele:
For reference:
# orphan rule
In Rust you cannot provide implementations of a trait for a struct unless you are either the crate that defines the struct, or the crate that defines the trait.

# Why not to implement `Into`
```rust
#![allow(unused_variables)]

#[derive(Clone, Copy)]
struct A(u32);
#[derive(Clone, Copy)]
struct B(u32);

impl Into<u32> for A { 
    fn into(self) -> u32 {
        self.0
    }   
}

impl From<B> for u32 {
    fn from(b: B) -> u32 {
        b.0 
    }   
}

fn main() {
    let a = A(42);
    let b = B(31);
    let val_a: u32 = a.into();
    // let val2 = <u32>::from(a); >> trait not implemented error
    let val_b_into: u32 = b.into();
    let val_b_from = <u32>::from(b);
}
```

PR #75, opened 2020-03-16
Subvariants

strom-und-spiele:
Acts on the discussions in https://github.com/stm32-rs/stm32f3xx-hal/pull/67


strom-und-spiele:
ah dang I messed up and added the wrong commits. will try to fix this later. I wanted to only commit https://github.com/stm32-rs/stm32f3xx-hal/pull/75/commits/d50bcf1ef2c2e6fec40b4d2e80178e9b5471fb45

strom-und-spiele:
Travis fails b.c. it's running `rustc […] --cfg 'feature="stm32f303"'` which is what this PR is about…

src/lib.rs
```diff
@@ -1,19 +1,22 @@
...
-        stm32f318
-        stm32f302
-        stm32f303
-        stm32f373
-        stm32f378
-        stm32f334
-        stm32f328
-        stm32f358
-        stm32f398"
+    "This crate requires you to specify your target device as a feature.
```
Sh3Rm4n: I do understand your intention. This would remove the work to update this file, if a new feature gets added. 

However, I'd like to keep the list. Or at least update it to the current list of features avaliable. 
I think it is more helpful, when rustc directly tells you, which feature to use, without having to lookup the README on github / docs.rs or crates.io. 

Suggesting that there are more details in the README is a nice addition though :) 

Thinking about it we don't even include the README in `Cargo.toml` yet and I like to see a feature explanation as a doc comment on top of `lib.rs`. But this can be implemented in a follow up PR :)

README.md
```diff
@@ -41,3 +41,38 @@ Almost all of the implementation was shamelessly adapted from the
 ## License
 
 [0-clause BSD license](LICENSE-0BSD.txt).
+
+## Selecting the right feature gate
```
Sh3Rm4n: Could you move this above the license notice?

README.md
```diff
@@ -41,3 +41,38 @@ Almost all of the implementation was shamelessly adapted from the
...
+
+If your code built perfectly before and throws a compile error now, don't worry.
+While there are a lot of cool new features coming your way, we want to make sure
+your old code still behaves the same and there are no wrong assumptions
+creating unexpected behaviour.
+
+As this crate is still under fundamental development, expect more sub-variants
+to pop up in the future.
+
+### Steps to select the right feature
```
Sh3Rm4n: In general, I think it would be more useful for the user, to not have to look at the `Cargo.toml` of this crate to find out what features are available. As a user, I want to only look at the source code as the last resort. 

Listing all available features (in this case all the sub-variants of the stm32f3xx family) explicitly in the README and maybe also in `lib.rs` would carry the information more efficiently towards the user.

Sh3Rm4n:
I really like the graphic included in the summary from [stm](https://www.st.com/en/microcontrollers-microprocessors/stm32f303.html) (middle of the page)

I would not include this image in the README, but maybe an explanation about the code?

There is a perfect graphic in the [datasheet](https://www.st.com/resource/en/datasheet/stm32f303cb.pdf) (page 142 section 8) Not every information is relevant. 

But on this part, I'm currently a bit confused. The Flash memory size letter is the one, which is referenced in the reference manual, on which distinction are made, which we replicate in the crate currently. 

But I would think that the pin count [C, R, V] does determine how many gpio banks are available. I've to think about this part a little bit more.

ci/build.py
```diff
@@ -28,8 +28,9 @@ def main():
     crate_info = cargo_meta["packages"][0]
 
     features = [
-        "{},rt".format(x) for x in crate_info["features"].keys()
-        if x not in ["device-selected", "rt", "default", "unproven"]
+        "{},rt".format(feature)
+        for feature, derived in crate_info["features"].items()
+        if "device-selected" in derived
```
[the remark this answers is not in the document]
Sh3Rm4n: Okay, i just checked the CI output, so it does seems to work :)

Cargo.toml
```diff
@@ -48,17 +48,22 @@ usbd-serial = "0.1.0"
 [features]
 default = ["unproven"]
 device-selected = []
+needs-subvariant = []
 rt = ["stm32f3/rt"]
-stm32f301 = ["stm32f3/stm32f301", "device-selected"]
+stm32f301 = ["stm32f3/stm32f301", "needs-subvariant"]
```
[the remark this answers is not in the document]
strom-und-spiele: Using the deprecated attribute seems to be impossible for features,

`needs-subvariant` allows for a kind of _deprecate compiler error_:
`device-selected` is only the opposide of `needs-subvariant` if the feature starts with `stm32f`. 
Using `needs-subvariant` allows to discriminate between a user not going into detail and a user not setting `stm32fxxx` at all.

src/lib.rs
strom-und-spiele: As much as I don't like to have information duplication (as it is a given that the information will diverge some day) I love the compiler telling me what to do to solve my problem.

@Sh3Rm4n 
> […] I like to see a feature explanation as a doc comment on top of `lib.rs`.

How will this be displayed in the docs?
And what is the benefit over just putting it into a compile error?

strom-und-spiele:
> There is a perfect graphic in the [datasheet](https://www.st.com/resource/en/datasheet/stm32f303cb.pdf) (page 142 section 8)[…]
> But on this part, I'm currently a bit confused. The Flash memory size letter is the one, which is referenced in the reference manual, on which distinction are made, which we replicate in the crate currently.

This list does not seem to be exhaustive. It only helps you specifying the STM32F303x[BCDE] type of chips. 
As far as I was involved, we discriminated between stm32f303x[BCDE] and stm32f303x[68].

README.md
strom-und-spiele: resolved in https://github.com/stm32-rs/stm32f3xx-hal/pull/75/commits/ced7a7ece4b4319ed3e03519fdf7ea19736cf5ed

Cargo.toml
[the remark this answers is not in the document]
Sh3Rm4n: That's true. But to make it effective as a deprecation, I would not change it to a compile error. 

Either keep the `needs-subvariant` feature, and spit out a less verbose warning (does `compiler_warn!` exist?), which states that the this feature is deprecated and choose a Variant, which has the least common features (e.g. `stm32f303x6` for the f303s). 
We could keep it that way for one or more version and than remove the intermediate step (soft breaking change)

Or remove the `needs-subvariant` feature (but this is not my favored way), which results into a compile error of the feature is still selected by the user (this would be a hard breaking change)

Either way, we should put a warning / breaking change note in the changelog.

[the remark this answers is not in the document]
strom-und-spiele: > Either keep the needs-subvariant feature, and spit out a less verbose warning (does compiler_warn! exist?),

Unfortunately, `compier_warn!` doesnt exist (see discussion [here](https://internals.rust-lang.org/t/pre-rfc-add-compile-warning-macro/9370) and links there) and I couldnt find a similar way of warning the user.

> choose a Variant, which has the least common features

Actually `stm32f303` would be the one with the least common features. Thats where all the confusion arises. So thats why I'm very much in favor of incrementing the version number.

README.md
```diff
@@ -38,6 +38,63 @@ Almost all of the implementation was shamelessly adapted from the
...
+*   stm32f373
+*   stm32f378
+*   stm32f334
+*   stm32f328
+*   stm32f358
+*   stm32f398
+
+### Background
+
+For some of the stm32f3xx chips there are sub-variants that behave differently
```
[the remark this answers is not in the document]
strom-und-spiele: https://github.com/stm32-rs/stm32f3xx-hal/pull/75/commits/3fd1bb8ccf0da25ebaa4dcb54da290ab8d0cb427

src/lib.rs
strom-und-spiele: nice. → https://github.com/stm32-rs/stm32f3xx-hal/pull/75/commits/ffa495a473a0769ee127ca8739ce30c6ccac9da6

Cargo.toml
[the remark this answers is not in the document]
Sh3Rm4n: Argh. So nothing we can do to cleanly deprecate this :/ 

So then, we should keep the `needs_subvariant` feature :)

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - Bump `stm32f3` dependency to `0.10.0` ([#42](https://github.com/stm32-rs/stm32f3xx-hal/pull/42))
+- The feature gate requires you to select a subvaraint if possible. ([#67](https://github.com/stm32-rs/stm32f3xx-hal/pull/67))
```
Sh3Rm4n, line 17: Note change as breaking, as stm32f303 feature won't compile anymore :) 

```suggestion
- **Breaking** The feature gate requires you to select a subvaraint if possible. ([#67](https://github.com/stm32-rs/stm32f3xx-hal/pull/67))
```

And this is out of date again (I feel like synching the CHANGELOG to master is tedious, but i feel like they won't be another release before this PR is merged :sweat_smile: )

src/lib.rs
```diff
@@ -1,19 +1,107 @@
...
+    *   stm32f398
+
+    Example: The STM32F3Discovery board has a STM32F303VCT6 chip.
+    So you want to expand your call to `cargo` with `--features stm32f303xc`.
+
+    For more information, see README -> Selecting the right feature gate.
+    "
+);
+
+#[cfg(all(not(feature = "device-selected"), feature = "needs-subvariant"))]
```
Sh3Rm4n, line 69: I still feel like this is to much duplication, when it only adds the extra information, that some features, specifically `stm32f303` will be removed. 

I would suggest changing this to: 

```rust
#[cfg(any(feature = "stm32f303", feature = "... other devices, which are no handled (this would be stm32f302 e.g."))]
compile_error!(
        "stm32f303 feature has been removed. Choose either stm32f303xc, ...xd, ... x6, ... x8"
);
```

And revert the change on the `cfg` above. 

As no other devices are affected yet, `any` is not needed for know. But soon https://github.com/stm32-rs/stm32f3xx-hal/pull/67 will add a reason to remove `stm32f302`. 

With this change, we can think about, if we even need `needs_subvariant`. But we don't have to remove it. We could also do:

```rust
#[cfg(feature = "needs-subvariant")]
compile_error!(
        "stm32f303 feature has been removed. Choose either stm32f303xc, ...xd, ... x6, ... x8"
);
```

src/lib.rs
```diff
@@ -1,19 +1,107 @@
...
+    Example: The STM32F3Discovery board has a STM32F303VCT6 chip.
+    So you want to expand your call to `cargo` with `--features stm32f303xc`.
+
+    For more information, see the [README](https://github.com/stm32-rs/stm32f3xx-hal/blob/master/README.md#selecting-the-right-chip)
+ */
 #![no_std]
 #![allow(non_camel_case_types)]
 
-#[cfg(not(feature = "device-selected"))]
+#[cfg(all(not(feature = "device-selected"), not(feature = "needs-subvariant")))]
```
Sh3Rm4n, line 36: Revert this. Currently, there is no device, which simultaneously has `device-selected` and `needs-subvariant` and it is also not planned in the future :)

README.md
```diff
@@ -38,6 +38,74 @@ Almost all of the implementation was shamelessly adapted from the
...
+## Selecting the right chip
+
+This crate requires you to specify your target chip as a feature.
+
+*Example: The STM32F3Discovery board has a STM32F303VCT6 chip.
+So you want to expand your call to `cargo` with `--features stm32f303xc`.*
+
+### Possible chips
+
+[comment]: # (Any changes here should be mirrored in src/lib.rs)
```
Sh3Rm4n, line 50: Ha nice trick about the comment. Wouldn't have come to my mind. HTML comments are also valid for markdown

```md
<!-- this is a comment -->
```

Cargo.toml
```diff
@@ -48,17 +48,23 @@ usbd-serial = "0.1.0"
...
 stm32f302 = ["stm32f3/stm32f302", "device-selected"]
-stm32f303 = ["stm32f3/stm32f303", "device-selected"]
-stm32f303xb = ["stm32f303", "stm32-usbd/ram_access_1x16"]
-stm32f303xc = ["stm32f303", "stm32-usbd/ram_access_1x16"]
-stm32f303xd = ["stm32f303", "stm32-usbd/ram_access_2x16"]
-stm32f303xe = ["stm32f303", "stm32-usbd/ram_access_2x16"]
-stm32f303x6 = ["stm32f303"]
-stm32f303x8 = ["stm32f303"]
+stm32f303 = ["stm32f3/stm32f303", "needs-subvariant"]
+stm32f303xb = ["stm32f303", "stm32-usbd/ram_access_1x16", "device-selected"]
```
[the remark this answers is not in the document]
strom-und-spiele: It's just that we are using the `#[cfg(any(feature = "stm32f303",…))]` in a lot of places. I thought it would be a little too verbose to use the `stm32f3/stm32f303` everywhere but I also see your point.

src/lib.rs
strom-und-spiele: I agree on the shorter error message. The rest depends on how we want to set up feature gates. maybe it's more efficient if we chat about it and come to a conclusion during that discussion.

src/lib.rs
strom-und-spiele: if you use `--features stm32f318` you only get `device-selected`.
if you use `--features stm32f303` you only get `needs-subvariant`.
if you use `--features stm32f303xc` you will get `device-selected` and `stm32f303`. by getting `stm32f303` you'll get `needs-subvariant` indirectly.
Also having neither `device-selected` nor `needs-subvariant` means you simply didnt specify any chip.
So all combinations are possible.

README.md
strom-und-spiele: tjhanks - I made the extra efford to search for it because I wanted pure markdown ;)

src/lib.rs
strom-und-spiele: I like the more specific error messages introduced in https://github.com/stm32-rs/stm32f3xx-hal/pull/75/commits/0fa4b138f50ea1610355066956c2b494b918aefb  more. hope I didn't go into the completely wrong direction with this ;)

Sh3Rm4n:
Thank you for this PR! :)

strom-und-spiele:
Should have "fixed" #81 by disallowing to select `stm32f303`

PR #80, opened 2020-03-23
Fixes to prevent panic setting up TIM2 in pwm example

jkboyce:
Hi, the pwm example causes a panic on TIM2 configuration, in the call to `tim2()`. Root cause is that the default 4 MHz SYSCLK isn't fast enough to deliver the resolution and frequency requested (TIM2 needs a clock of at least 160_000 * 50 Hz = 8 MHz, and this can't be greater than SYSCLK). The actual panic occurs [here](https://github.com/stm32-rs/stm32f3xx-hal/blob/master/src/pwm.rs#L468) when `prescale_factor=0`. The easy fix is to bump SYSCLK to at least 8 MHz.

I also removed PB3 and PB7 from the set of output pins, since on the F3 Discovery board these (may) be used for other things. PB3 may be attached to the SWO line of the debugger port (if solder bridge SB10 is on), and monkeying with that pin kills ITM debugging. PB7 is attached to the I2C1 SDA line of the LSM303DLHC accelerometer. Anyway both outputs are mirrored on other pins so removing them doesn't materially change the example.

### DIFF
--- a/examples/pwm.rs
+++ b/examples/pwm.rs
@@ -23,7 +23,7 @@ fn main() -> ! {
     // Configure our clocks
     let mut flash = dp.FLASH.constrain();
     let mut rcc = dp.RCC.constrain();
-    let clocks = rcc.cfgr.freeze(&mut flash.acr);
+    let clocks = rcc.cfgr.sysclk(16.mhz()).freeze(&mut flash.acr);
 
     // Prep the pins we need in their correct alternate function
     let mut gpioa = dp.GPIOA.split(&mut rcc.ahb);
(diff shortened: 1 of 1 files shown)


Sh3Rm4n:
Thanks for the PR 👍 

> Root cause is that the default 4 MHz SYSCLK isn't fast enough to deliver the resolution and frequency requested. 

This seems like a bug, that i'm trying to to fix (https://github.com/stm32-rs/stm32f3xx-hal/pull/67)

Ah the panic is happening because of an underflow. Yeah this makes sense. And the pin remapping is much appreacheated. 

LGTM, but maybe @IamfromSpace could take another look, if he wants :)

examples/pwm.rs
```diff
@@ -23,7 +23,7 @@ fn main() -> ! {
     // Configure our clocks
     let mut flash = dp.FLASH.constrain();
     let mut rcc = dp.RCC.constrain();
-    let clocks = rcc.cfgr.freeze(&mut flash.acr);
+    let clocks = rcc.cfgr.sysclk(16.mhz()).freeze(&mut flash.acr);
```
Sh3Rm4n: I like the change, as being more explicit about the clock configuration for this example makes sense. 
It wouldn't be needed, if the bug I mentioned above would not exist, but this does not hurt.

PR #64, opened 2020-03-07
Update dependencies

Sh3Rm4n:

PR #85, opened 2020-04-04
Clean up dependency versions in Cargo.toml

ra-kete:
This PR changes the version specifications in the Cargo.toml:
 - `"0.y.z"` -> `"0.y"`
 - `"x.y.z"` -> `"x"`

This does not functionally change anything, since version strings
without any operators default to the caret requirement anyway ([see Cargo docs][1]).
This changes just makes it obvious.

[1]: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,32 +19,32 @@ targets = ["thumbv7em-none-eabihf"]
 travis-ci = { repository = "stm32-rs/stm32f3xx-hal" }
 
 [dependencies]
-cortex-m = ">=0.5.8,<0.7"
-cortex-m-rt = "0.6.8"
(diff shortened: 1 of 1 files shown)


ra-kete:
I just noticed that this PR conflicts with #64.

AFAICT #64 actually is not necessary to update the dependencies, due to the implicit caret requirement mentioned above. Which just goes to show that the way dependencies are specified now is confusing.

Sh3Rm4n:
I like your approach better than mine in #64. 

I know that updating minor versions is not really necessary, that's why I hesitated to merge #64. 

I'll close it in favour of yours.

Issue #66, opened 2020-03-08
Support DMA


Sh3Rm4n:
Direct Memory Access allows peripherals to write to memory directly, without needing buffers. This saves memory and is a handy feature.

ra-kete:
I'm interested in this too. The project I'm working on pretty much demands DMA (in tandem with ADC and SPI). That's a major reason I don't use this HAL for it but wrote [my own one](https://gitlab.com/ra_kete/stm32f303xc-hal), based on [japari's blog post](https://blog.japaric.io/safe-dma/). I planned to upstream that at some point, but wasn't really sure how to go about it. My implementation is pretty specific to my needs and probably needs to be refactored a bit to be generally useful. And of course it supports only STM303xC MCUs.

I'd be willing to try my hand on implementing something that's useful for everyone. If you don't already have other plans, of course :)

Sh3Rm4n:
Cool. I have no plans to implement this myself atm. So I'm looking forward for any support on this. 

I'm okay with it being only compatible with the stm32303xc for now (similar to https://github.com/stm32-rs/stm32f3xx-hal/pull/47) If it is working, we can always incrementally add support for more stm32f3xx chips later in followup PRs. :)

ra-kete:
I've created a PR that adds MVP DMA support, i.e. the minimum amount of functionality that's useful to actually do something. I think it's a good idea to add more functionality incrementally instead of having one huge PR in the end that nobody understands and that takes forever to get merged.

Adding the missing DMA features I can think of:

- [x] generic support for one-shot DMAs
- [ ] generic support for circular DMAs
- [x] support for serial DMA
- [ ] support for SPI DMA
- [ ] support for I2C DMA
- [ ] support for ADC DMA (blocked by #20)

ra-kete:
Here are a few references I found helpful for DMA implementation:

- [The Embedonomicon on DMA](https://docs.rust-embedded.org/embedonomicon/dma.html)
- [`embedded-hal` issue about DMA](https://github.com/rust-embedded/embedded-hal/issues/37)
- [DMA Implementation in `stm32l0xx-hal`](https://github.com/stm32-rs/stm32l0xx-hal/blob/master/src/dma.rs)

PR #86, opened 2020-04-05
Initial DMA support

ra-kete:
This PR is the first part of an effort to add DMA support to this crate (see issue #66). It adds initial support for DMA, including:

- A `DmaExt` extension trait that splits a DMA peripheral into individual channels
- A `Transfer` abstraction that provides memory-safe one-shot DMA transfers
- New methods on serial receivers and transmitters to read/write data via DMA

This works only on STM32F303 devices for now.

### DIFF
--- a/src/dma.rs
+++ b/src/dma.rs
@@ -39,7 +39,9 @@ impl<B, C: Channel, T> Transfer<B, C, T> {
         B: WriteBuffer + 'static,
         T: Target<C>,
     {
-        let (ptr, len) = buffer.write_buffer();
+        // NOTE(unsafe) cannot call `&mut self` methods on `buffer` because its
+        // concrete type is unknown here
+        let (ptr, len) = unsafe { buffer.write_buffer() };
 
         channel.set_memory_address(ptr as u32, Increment::Enable);
         channel.set_transfer_length(len);
@@ -55,7 +57,9 @@ impl<B, C: Channel, T> Transfer<B, C, T> {
         B: ReadBuffer + 'static,
         T: Target<C>,
     {
-        let (ptr, len) = buffer.read_buffer();
+        // NOTE(unsafe) cannot call `&mut self` methods on `buffer` because its
+        // concrete type is unknown here
+        let (ptr, len) = unsafe { buffer.read_buffer() };
 
         channel.set_memory_address(ptr as u32, Increment::Enable);
         channel.set_transfer_length(len);
@@ -142,7 +146,11 @@ impl<B, C: Channel, T> TransferInner<B, C, T> {
 /// The implementing type must be safe to use for DMA reads. This means:
 ///
 /// - It must be a pointer that references the actual buffer.
-/// - The requirements documented on `read_buffer` must be fulfilled.
+/// - The following requirements must be fulfilled by `read_buffer`:
+///   - The function must always return the same values, if called multiple
+///     times.
+///   - The memory specified by the returned pointer and size must not be
+///     freed as long as `self` is not dropped.
 pub unsafe trait ReadBuffer {
     type Word;
 
@@ -155,11 +163,9 @@ pub unsafe trait ReadBuffer {
     ///
     /// # Safety
     ///
-    /// - This function must always return the same values, if called multiple
-    ///   times.
-    /// - The memory specified by the returned pointer and size must not be
-    ///   freed as long as `self` is not dropped.
-    fn read_buffer(&self) -> (*const Self::Word, usize);
+    /// Once this method has been called, it is unsafe to call any `&mut self`
+    /// methods on this object as long as the returned value is used (for DMA).
+    unsafe fn read_buffer(&self) -> (*const Self::Word, usize);
 }
 
 /// Trait for buffers that can be given to DMA for writing.
@@ -170,7 +176,11 @@ pub unsafe trait ReadBuffer {
 ///
 /// - It must be a pointer that references the actual buffer.
 /// - `Target` must be a type that is valid for any possible byte pattern.
-/// - The requirements documented on `write_buffer` must be fulfilled.
+/// - The following requirements must be fulfilled by `write_buffer`:
+///   - The function must always return the same values, if called multiple
+///     times.
+///   - The memory specified by the returned pointer and size must not be
+///     freed as long as `self` is not dropped.
 pub unsafe trait WriteBuffer {
     type Word;
 
@@ -183,11 +193,9 @@ pub unsafe trait WriteBuffer {
     ///
     /// # Safety
     ///
-    /// - This function must always return the same values, if called multiple
-    ///   times.
-    /// - The memory specified by the returned pointer and size must not be
-    ///   freed as long as `self` is not dropped.
-    fn write_buffer(&mut self) -> (*mut Self::Word, usize);
+    /// Once this method has been called, it is unsafe to call any `&mut self`
+    /// methods on this object as long as the returned value is used (for DMA)..
+    unsafe fn write_buffer(&mut self) -> (*mut Self::Word, usize);
 }
 
 // Blanked implementations for common DMA buffer types.
@@ -199,7 +207,7 @@ where
 {
     type Word = T::Word;
 
-    fn read_buffer(&self) -> (*const Self::Word, usize) {
+    unsafe fn read_buffer(&self) -> (*const Self::Word, usize) {
         self.as_read_buffer()
     }
 }
@@ -211,7 +219,7 @@ where
 {
     type Word = T::Word;
 
-    fn write_buffer(&mut self) -> (*mut Self::Word, usize) {
+    unsafe fn write_buffer(&mut self) -> (*mut Self::Word, usize) {
         self.as_write_buffer()
     }
 }
--- a/src/dma.rs
+++ b/src/dma.rs
@@ -52,8 +52,10 @@ impl<B, C: Channel, T: Target> Transfer<B, C, T> {
         B: WriteBuffer + 'static,
         T: OnChannel<C>,
     {
-        // NOTE(unsafe) cannot call `&mut self` methods on `buffer` because its
-        // concrete type is unknown here
+        // NOTE(unsafe) We don't know the concrete type of `buffer` here, all
+        // we can use are its `WriteBuffer` methods. Hence the only `&mut self`
+        // method we can call is `write_buffer`, which is allowed by
+        // `WriteBuffer`'s safety requirements.
         let (ptr, len) = unsafe { buffer.write_buffer() };
         let len = u16(len).expect("buffer is too large");
 
@@ -75,8 +77,10 @@ impl<B, C: Channel, T: Target> Transfer<B, C, T> {
         B: ReadBuffer + 'static,
         T: OnChannel<C>,
     {
-        // NOTE(unsafe) cannot call `&mut self` methods on `buffer` because its
-        // concrete type is unknown here
+        // NOTE(unsafe) We don't know the concrete type of `buffer` here, all
+        // we can use are its `ReadBuffer` methods. Hence there are no
+        // `&mut self` methods we can call, so we are safe according to
+        // `ReadBuffer`'s safety requirements.
         let (ptr, len) = unsafe { buffer.read_buffer() };
         let len = u16(len).expect("buffer is too large");
 
@@ -168,11 +172,11 @@ impl<B, C: Channel, T: Target> TransferInner<B, C, T> {
 /// The implementing type must be safe to use for DMA reads. This means:
 ///
 /// - It must be a pointer that references the actual buffer.
-/// - The following requirements must be fulfilled by `read_buffer`:
-///   - The function must always return the same values, if called multiple
+/// - As long as no `&mut self` method is called on the implementing object:
+///   - `read_buffer` must always return the same value, if called multiple
 ///     times.
-///   - The memory specified by the returned pointer and size must not be
-///     freed as long as `self` is not dropped.
+///   - The memory specified by the pointer and size returned by `read_buffer`
+///     must not be freed as long as `self` is not dropped.
 pub unsafe trait ReadBuffer {
     type Word;
 
@@ -186,7 +190,7 @@ pub unsafe trait ReadBuffer {
     /// # Safety
     ///
     /// Once this method has been called, it is unsafe to call any `&mut self`
-    /// methods on this object as long as the returned value is used (for DMA).
+    /// methods on this object as long as the returned value is in use (by DMA).
     unsafe fn read_buffer(&self) -> (*const Self::Word, usize);
 }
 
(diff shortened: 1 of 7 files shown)


src/serial.rs
```diff
@@ -36,24 +23,20 @@ use crate::gpio::gpioc::{PC10, PC11, PC4, PC5};
...
+        ops::{Deref, DerefMut},
+        pin::Pin,
+    };
+}
 
-use crate::gpio::AF7;
-use crate::rcc::{Clocks, APB1, APB2};
-use crate::time::Bps;
+#[cfg(feature = "stm32f303")]
+use dma_imports::*;
```
ra-kete, line 39: I cleaned up the imports a bit while I added my own ones. I noticed that model-specific imports add a lot of duplicated `cfg` statements. It would help readability a lot if we could reduce this duplication.

According to [this SO answer](https://stackoverflow.com/questions/43070430/is-there-a-way-to-use-the-cfgfeature-check-on-multiple-statements), we can solve this two ways:

1. Using an ad-hoc submodule and then re-exporting its members in the parent. This is the technique you see here.
2. Using the [cfg-if](https://crates.io/crates/cfg-if) crate.

I'm not too fond of the submodule solution, as it still adds unnecessary visual noise and still needs the `cfg` to be repeated twice. The `cfg-if` solution looks better to me, but would require an additional dependency.

src/serial.rs
```diff
@@ -373,7 +369,67 @@ macro_rules! hal {
                 }
             }
 
-            impl Default<u8> for Tx<$USARTX> {}
+            impl blocking::serial::write::Default<u8> for Tx<$USARTX> {}
+
+            #[cfg(feature = "stm32f303")]
+            impl Rx<$USARTX> {
+                /// Fill the buffer with received data using DMA.
+                pub fn read_all<B, C>(
```
ra-kete, line 377: I'll rename that to `read_exact`, that's more fitting.

Sh3Rm4n:
Thank you very much for this PR and providing so much extra information in https://github.com/stm32-rs/stm32f3xx-hal/issues/66. Now I have enough information to read into. I'll try to review this in the next days :)

src/serial.rs
Sh3Rm4n: I welcome this change :+1:

Sh3Rm4n:
Am I right, that this is on hold as long as [this](https://github.com/ra-kete/dma-poc) is still in progress? :)

ra-kete:
Not necessarily. The final goal of this project is to update the Embedonomicon, we don't need to wait for that. I think the discussion around the buffer type is most relevant for this PR and so far as I'm aware, the current proposal is sound. At least nobody has raised any issues with it yet.

So I would suggest I update this PR to make use of the suggested traits and then we try to get it merged. That way we can also get some practical experience with these traits, which we should before we recommend them for everyone.

ra-kete:
I'd say it's ready for review now.

ra-kete:
Hey @Sh3Rm4n, I don't want to be a bother but that has been waiting for a while. Anything I can do to help move it forward?

src/serial.rs
```diff
@@ -251,21 +237,21 @@ macro_rules! hal {
...
-                    // UE: enable USART
-                    // RE: enable receiver
-                    // TE: enable transceiver
-                    usart
-                        .cr1
-                        .modify(|_, w| w.ue().set_bit().re().set_bit().te().set_bit());
+                    usart.cr3.modify(|_, w| {
+                        w.dmar().enabled();  // enable DMA for reception
+                        w.dmat().enabled()   // enable DMA for transmission
+                    });
```
thalesfragoso, line 248: Hmm, maybe we should only enable the DMA when actually using it ?

thalesfragoso:
Oh, I forgot something, maybe we should have a method on the `Target` trait to get the peripheral address, that way we can set the address on the DMA code, what do you think ? Seems less error prone.

src/dma.rs
```diff
@@ -0,0 +1,699 @@
...
+    buffer: B,
+    channel: C,
+    target: T,
+}
+
+impl<B, C: Channel, T> TransferInner<B, C, T> {
+    /// Stop this transfer
+    fn stop(&mut self) {
+        self.channel.disable();
+        atomic::compiler_fence(Ordering::Acquire);
```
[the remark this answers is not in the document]
ra-kete: But the `self.channel.disable()` does a read, no? It is implemented as:
```rust
self.ch().cr.modify(|_, w| w.en().disabled());
```
And `modify` does a read-modify-write as far as I understand. I mean, it has to. How else would it set a single bit and leave the others unchanged?

src/dma.rs
```diff
@@ -0,0 +1,699 @@
...
+    /// - pointer to the start of the buffer
+    /// - buffer size in words
+    ///
+    /// # Safety
+    ///
+    /// - This function must always return the same values, if called multiple
+    ///   times.
+    /// - The memory specified by the returned pointer and size must not be
+    ///   freed as long as `self` is not dropped.
+    fn read_buffer(&self) -> (*const Self::Word, usize);
```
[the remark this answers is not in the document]
ra-kete: Well, the `# Safety` comment affects implementation only, not calling the function, so it is actually tied to the `unsafe` of the whole trait. But you are right, it needs to be in the `# Safety` section of the trait then too.

src/dma.rs
```diff
@@ -0,0 +1,699 @@
...
+                        use Event::*;
+
+                        // NOTE(unsafe) atomic write to a stateless register
+                        unsafe {
+                            &(*$DMAx::ptr()).ifcr.write(|w| match event {
+                                HalfTransfer => w.$chtifi().set_bit(),
+                                TransferComplete => w.$ctcifi().set_bit(),
+                                TransferError => w.$cteifi().set_bit(),
+                                Any => w.$cgifi().set_bit(),
+                            });
```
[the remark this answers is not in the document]
ra-kete: I tend to agree. But what about the scenario where both `TransferComplete` and `HalfTransfer` occured and the user clears `TransferComplete`? In this case, wouldn't they expect a call with `Any` to still return `true`, since the `HalfTransfer` is still set?

IMO, we should only clear the global flag when all other flags are cleared too. Or just leave it as is and document that the user has to take care of clearing the global flag.

src/dma.rs
```diff
@@ -0,0 +1,699 @@
...
+
+                impl private::Channel for $Ci {
+                    fn ch(&self) -> &stm32::dma1::CH {
+                        // NOTE(unsafe) $Ci grants exclusive access to this register
+                        unsafe { &(*$DMAx::ptr()).$chi }
+                    }
+                }
+
+                impl Channel for $Ci {
+                    fn event_occured(&self, event: Event) -> bool {
```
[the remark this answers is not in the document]
ra-kete: Ups, ignored the spellcheck here :)

src/dma.rs
```diff
@@ -0,0 +1,699 @@
+//! Direct memory access (DMA) controller
+
+// Currently DMA is only supported for STM32F303 MCUs.
+// Remove these `allow`s once all models have support.
+#![cfg_attr(not(feature = "stm32f303"), allow(unused_imports, unused_macros))]
```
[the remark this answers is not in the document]
ra-kete: I remember having some issues with conditionally adding the `dma` module itself, but can't think of why anymore. I'll test that again.

src/serial.rs
ra-kete: Here I just went with what the stm32l0xx-hal is doing: https://github.com/stm32-rs/stm32l0xx-hal/blob/master/src/serial.rs#L353

I don't know what their reasoning was, but for me it boils down to:
- Setting these bits does not start any additional peripherals, so it *should* not lead to more resource usage, so we should be fine to always do it. But to be honest I don't know what setting the bits really does on a hardware level.
- We could set these bits inside `read_exact` and `write_all` instead, but that would mean unnecessary register modifies when those methods are called multiple times.

But it probably doesn't matter much in either case. I'd not be opposed to moving these DMA enables if you think it would be better that way.

ra-kete:
> Oh, I forgot something, maybe we should have a method on the `Target` trait to get the peripheral address, that way we can set the address on the DMA code, what do you think ? Seems less error prone.

I looked into that, but it's not easy to implement nicely. The naive transformation would yield something like:
```rust
/// Trait for defining DMA targets and their channels
pub unsafe trait Target<C: Channel> {
    fn peripheral_address() -> u32;
}

macro_rules! targets {
    (
        $dma:ident,
        $( $target:ty => ($C:ident, $reg:expr), )+
    ) => {
        $(
            unsafe impl Target<$dma::$C> for $target {
                fn peripheral_address() -> u32 {
                    unsafe { &$reg as *const _ as u32 }
                }
            }
        )+
    };
}

#[cfg(feature = "stm32f303")]
targets!(dma1,
    serial::Rx<stm32::USART1> => (C5, (*stm32::USART1::ptr()).rdr),
    serial::Tx<stm32::USART1> => (C4, (*stm32::USART1::ptr()).tdr),
    serial::Rx<stm32::USART2> => (C6, (*stm32::USART2::ptr()).rdr),
    serial::Tx<stm32::USART2> => (C7, (*stm32::USART2::ptr()).tdr),
    serial::Rx<stm32::USART3> => (C3, (*stm32::USART3::ptr()).rdr),
    serial::Tx<stm32::USART3> => (C2, (*stm32::USART3::ptr()).tdr),
);
```

Which I don't like much, since (aside from being ugly) I don't think the DMA module should concern itself with the details of specific USART registers. Such low-level knowledge is better kept in the serial module IMO is. Now, this could be resolved with an additional `PeripheralAddress` trait that's impl'd on `Rx` and `Tx` inside the serial module. However, I'm not sure the additional complexity is worth it, at least for the MVP.

I think at some point, when DMA is supported for more than just serial, we'll want to move more setup logic into the `Transfer::start_*` methods, to remove code duplication. But I'd like to wait until then to see what the actual common usage patterns are. Trying to predict that now feels like over-engineering.

src/dma.rs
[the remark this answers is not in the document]
thalesfragoso: The purpose of this fence is to prevent operations on the buffer from getting moved to before the DMA is disabled, that happens in the write part. In `modify`, the read comes before the write, so the `Acquire` fence is, effectively, between the read and write. That way, it prevents the operations from being moved to before the read, but not the write.

src/dma.rs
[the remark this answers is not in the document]
thalesfragoso: Hmm, makes sense, I think the best solution is to leave as is and have a good documentation about this then.

src/serial.rs
thalesfragoso: I think it starts sending DMA requests, not sure what implications this has on a disabled channel, specifically, if a old request would stay there until the DMA is enabled or if it would be instantly dropped.

thalesfragoso:
> I think at some point, when DMA is supported for more than just serial, we'll want to move more setup logic into the Transfer::start_* methods, to remove code duplication. But I'd like to wait until then to see what the actual common usage patterns are. Trying to predict that now feels like over-engineering.

Agreed.

src/dma.rs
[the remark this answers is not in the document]
ra-kete: You are right of course.

What do you think of adding an `unsafe { ptr::read_volatile(&0); }` after the `channel.disable()`? That's what suggested in the Embedonomicon to force a memory read. That way we still have a single fence for DMA start and a single fence for DMA stop, which I think is less confusing than having multiple ones. Also we'd avoid the `SeqCst` fence, which sounds more expensive to me.

src/serial.rs
ra-kete: Okay, "not sure what would happen" is a good reason to err on the safe side. I'll move those enables.

src/dma.rs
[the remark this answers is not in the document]
thalesfragoso: To be honest, I don't think that `SeqCst` would have a noticeable difference (in performance) from `Acquire` here in specific. And I'm almost sure that pushing a variable to the stack, taking its address and reading it would probably be more expensive, so I would go with `SeqCst` which is even more readable than having an `unsafe { ptr::read_volatile(&0); }`.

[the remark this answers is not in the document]
ra-kete: You're right, I'm getting lost in premature optimization. Still I think it would be preferable to just change this fence here to `SeqCst` ordering instead of introducing a second one and making it harder to be sure that we have a fence in every relevant path.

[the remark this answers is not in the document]
thalesfragoso: Makes sense.

src/serial.rs
ra-kete: Implementing this has been a bit more work than I expected, because if we only enable DMA when we actually start a transfer, we should also disable it again when the transfer is over. So I repurposed the `Target` trait to provide `enable_dma` and `disable_dma` methods that are now used as callbacks when a `Transfer` starts and stops. The mapping from targets to DMA channels is split into a separate trait called `OnChannel` now.

src/dma.rs
```diff
@@ -39,7 +39,9 @@ impl<B, C: Channel, T> Transfer<B, C, T> {
         B: WriteBuffer + 'static,
         T: Target<C>,
     {
-        let (ptr, len) = buffer.write_buffer();
+        // NOTE(unsafe) cannot call `&mut self` methods on `buffer` because its
+        // concrete type is unknown here
```
thalesfragoso, line 56:
```suggestion
        // NOTE(unsafe) We can't (and won't) call `&mut self` methods on
        // `buffer` after this call and until the end of the transfer, because
        // its concretetype is unknown here
```

thalesfragoso: Not sure about the wording, but we should explain this better.

src/dma.rs
```diff
@@ -55,7 +57,9 @@ impl<B, C: Channel, T> Transfer<B, C, T> {
         B: ReadBuffer + 'static,
         T: Target<C>,
     {
-        let (ptr, len) = buffer.read_buffer();
+        // NOTE(unsafe) cannot call `&mut self` methods on `buffer` because its
+        // concrete type is unknown here
```
thalesfragoso, line 79:
```suggestion
        // NOTE(unsafe) We can't (and won't) call `&mut self` methods on
        // `buffer` after this call and until the end of the transfer, because
        // its concretetype is unknown here
```

src/dma.rs
```diff
@@ -142,7 +146,11 @@ impl<B, C: Channel, T> TransferInner<B, C, T> {
 /// The implementing type must be safe to use for DMA reads. This means:
 ///
 /// - It must be a pointer that references the actual buffer.
-/// - The requirements documented on `read_buffer` must be fulfilled.
+/// - The following requirements must be fulfilled by `read_buffer`:
+///   - The function must always return the same values, if called multiple
+///     times.
```
thalesfragoso, line 173: I don't think this is true in the usual `StableDeref` sense of things.

src/dma.rs
```diff
@@ -170,7 +176,11 @@ pub unsafe trait ReadBuffer {
 ///
 /// - It must be a pointer that references the actual buffer.
 /// - `Target` must be a type that is valid for any possible byte pattern.
-/// - The requirements documented on `write_buffer` must be fulfilled.
+/// - The following requirements must be fulfilled by `write_buffer`:
+///   - The function must always return the same values, if called multiple
+///     times.
```
thalesfragoso, line 203: Same as above.

src/dma.rs
ra-kete: I thought it was clear, but I'm probably in too deep :) I'll update this with more words.

src/dma.rs
ra-kete: Yeah, I also had issues with that. The `{Read,Write}Buffers` demand that the buffer we give out remains stable and valid as long as the object lives. But that is not true, when we use `StableDeref`. With `StableDeref` it is only true if we also demand that `&mut self` is not called.

Ideally, we would just drop the use of `StableDeref` and be strict on the initial demand. That would keep the safety guarantees relatively simple with few edge cases. Also the `{read,write}_buffer` methods wouldn't need to be `unsafe` anymore. BUT that would also prevent us from using `Vec` and `Box` for DMA. Those could only safely implement the traits if they were wrapped in a newtype that makes the `&mut self` methods inaccessible. Which is inconvenient. And shouldn't be necessary, since `Vec` and `Box` are perfectly fine for DMA...

So... I'll update those `Safety` comments to also include the `&mut self` edge case. But maybe we can think of something better at some point.

Issue #84, opened 2020-04-02
Implement InputPin trait for PA1<Output<OpenDrain>>


peter-wilkins:
I'm trying to read a dht11 humidity sensor with a stm32f303 discovery board

the example in the dht11 crate is for stmf0xx boards:
https://github.com/michaelbeaumont/dht-sensor/blob/master/examples/stm32f042.rs#L30

Rather than give in and just buy an stm32f0xx board I'd like to have a crack at adding this trait. If anyone knows that it is futile (eg  `Output<OpenDrain>` is not both `InputPin` and `OutputPin`  for the f3) or not useful in anyway please say so.

Part-converted example: https://gist.github.com/peter-wilkins/b5c914b1f8f5b4098b198133b4280472

I'm just starting out with embedded but will attempt to put together an PR, it may take some time. Any advice appreciated.

Sh3Rm4n:
The hardware design of the gpio's from the stm32fx-line should all be the same, so nothing speaks against adding the `InputPin` trait to `Output<OpenDrain>`. And it is clearly missing right now in this HAL. Seems like that besides stm32f0xx, stm32f1xx and stm32f4xx have these traits as well. 

So if you are willing to put together a PR, go for it. I appreciate every contribution. :) 

You could take the implementation from the [stm32f4xx-hal](https://github.com/stm32-rs/stm32f4xx-hal/blob/ddccb4fc48908f66f3848e686e87961a0df73a43/src/gpio.rs#L699). 

[stm32f1xx-hal](https://github.com/stm32-rs/stm32f1xx-hal/blob/master/src/gpio.rs) has a similar solution, but only implements it for OpenDrain. I don't know if we should limit the Trait to `Output<OpenDrain>`. I tend towards the more generic solution from the stm32f4xx-hal :)

David-OConnor:
I think I'm running into this as well, in an attempt to get the DS18B20 temp sensor working.

David-OConnor:
https://github.com/stm32-rs/stm32f3xx-hal/pull/105

therealprof:
> You could take the implementation from the [stm32f4xx-hal](https://github.com/stm32-rs/stm32f4xx-hal/blob/ddccb4fc48908f66f3848e686e87961a0df73a43/src/gpio.rs#L699).

[...]

Issue #37, opened 2019-12-06
Registers with "Global" Bits (AFRL/MODER/APB1/etc) _must_ not be Exposed


IamfromSpace:
# Summary

The current way we go about this is both un-ergonomic and unsound.

## Current Usage

Today, the strategy for handling registers like these is to allow only the HAL to mutate them (via `pub(crate)`), and then assure that they are available for mutation from the user.

For example if we want to put PA4 into an alternate function mode 2:
```
...
let mut gpioa = dp.GPIOA.split(&mut rcc.ahb);
gpioa.pa4.into_af2(&mut gpioa.moder, &mut gpioa.afrl);
```

The `AHB`, `MODER`, and `AFRL` have no public API, so they cannot be mutated directly, but they must be (mutably) passed into the `split` or `into_af2` so the HAL can modify them and correctly configure the pin.

## The Problem

Originally, my thought was a recommendation to improve ergonomics and hide more internal details from users.  However, I realized that this strategy is not actually sound.  The former leads to the latter, so we'll explore them in that order.

### Ergonomics

If we do not return or require these registers it greatly simplifies the API for users.  New users who are less familiar to MCU will be able to work in more concrete topics like peripherals, rather than registers.

Also, composition becomes much simpler.  Now, peripherals that accept pins must either a) require the pin in an already correctly configured state b) require registers that are not related its own function so it can configure the pin.

The former requires the user know extra details, and the latter is a break of the Law of Demeter (reaching into a dependency).

If instead moving a pin into alternate function 2 was as simple as:
```
let mut gpioa = dp.GPIOA.split();
gpioa.pa4.into_af2();
```

Then a peripheral could simply accept an unconfigured pin, _and the user would never even know that an alternate function was required_, further hiding details and simplifying user's code.  HAL's across multiple chips and manufactures would look more similar and wouldn't require re-learning the nuances of each chip.

The two reasons that this hasn't been done before is likely not wanting to mask these details and the number of resulting `unsafe` blocks.

It is not the job of a HAL to _teach_ a user how their chip works.  There should absolutely be resources to enable the curious!  But an _abstraction_ layer is about hiding. And the resulting `unsafe` blocks are not any _less_ safe than today, due to soundness issues.

### Unsoundness

This strategy has been built on the premise that:  If HAL users do not write unsafe code, and HAL contributors do not write unsafe code, then there will not be "unsafe" results.  This is not true.

As a contributor I can write the following "safe" function:
```
[...]

PR #88, opened 2020-04-06
WIP: add InputPin trait to output pins

peter-wilkins:
Closes https://github.com/stm32-rs/stm32f3xx-hal/issues/84

This builds and runs (I'm not getting a reading from sensor but that is probably some other issue like my wiring)

Out of curiosity, I've updated into_open_drain_output with the api from stm32f4 library which is simpler imo. Aware this is a breaking change so should probably not do this :)

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -546,26 +546,22 @@ macro_rules! gpio {
                         }
 
                         /// Configures the pin to operate as an open drain output pin
-                        pub fn into_open_drain_output(
-                            self,
-                            moder: &mut MODER,
-                            otyper: &mut OTYPER,
-                        ) -> $PXi<Output<OpenDrain>> {
-                            let offset = 2 * $i;
-
-                            // general purpose output mode
-                            let mode = 0b01;
-                            moder.moder().modify(|r, w| unsafe {
-                                w.bits((r.bits() & !(0b11 << offset)) | (mode << offset))
+                        pub fn into_open_drain_output(self) -> $PXi<Output<OpenDrain>> {
+                        let offset = 2 * $i;
+                        unsafe {
+                            &(*$GPIOX::ptr()).pupdr.modify(|r, w| {
+                                w.bits((r.bits() & !(0b11 << offset)) | (0b00 << offset))
                             });
+                            &(*$GPIOX::ptr()).otyper.modify(|r, w| {
+                                w.bits(r.bits() | (0b1 << $i))
+                            });
+                            &(*$GPIOX::ptr()).moder.modify(|r, w| {
+                                w.bits((r.bits() & !(0b11 << offset)) | (0b01 << offset))
+                            })
+                        };
 
-                            // open drain output
-                            otyper
-                                .otyper()
-                                .modify(|r, w| unsafe { w.bits(r.bits() | (0b1 << $i)) });
-
-                            $PXi { _mode: PhantomData }
-                        }
+                        $PXi { _mode: PhantomData }
+                    }
 
                         /// Configures the pin to operate as an push pull output pin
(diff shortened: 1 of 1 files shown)


src/gpio.rs
```diff
@@ -699,6 +695,19 @@ macro_rules! gpio {
 
                     #[cfg(feature = "unproven")]
                     impl<MODE> toggleable::Default for $PXi<Output<MODE>> {}
+
+                    impl<MODE> InputPin for $PXi<Output<MODE>> {
```
Sh3Rm4n, line 699: Almost done, this also has to be implemented at line 231 for `PXx` which is a "fully reased pin", line 425 `$PXx`, which is a partially erase pin. :) The implementation is almost the same. The compiler will scream the difference to you, or you could look at the other implementations :wink:

src/gpio.rs
```diff
@@ -546,26 +546,22 @@ macro_rules! gpio {
...
-                            moder: &mut MODER,
-                            otyper: &mut OTYPER,
-                        ) -> $PXi<Output<OpenDrain>> {
-                            let offset = 2 * $i;
-
-                            // general purpose output mode
-                            let mode = 0b01;
-                            moder.moder().modify(|r, w| unsafe {
-                                w.bits((r.bits() & !(0b11 << offset)) | (mode << offset))
+                        pub fn into_open_drain_output(self) -> $PXi<Output<OpenDrain>> {
```
Sh3Rm4n, line 549: Breaking changes in itself are fine. But changing only this, would mean other parts which take the `moder` and `otyper` register must be changed as well, to make it coherent. 

We had a discussion about this a while ago in https://github.com/stm32-rs/stm32f3xx-hal/issues/37. I've no opinion on this, but I'd rather tackle that as whole in another PR :)

src/gpio.rs
```diff
@@ -699,6 +695,19 @@ macro_rules! gpio {
...
+                    fn is_high(&self) -> Result<bool, Self::Error> {
+                        self.is_low().map(|v| !v)
+                    }
+
+                    fn is_low(&self) -> Result<bool, Self::Error> {
+                        // NOTE(unsafe) atomic read with no side effects
+                        Ok(unsafe { (*$GPIOX::ptr()).idr.read().bits() & (1 << $i) == 0 })
+                    }
+                }
                 )+
```
Sh3Rm4n, line 711:
```suggestion
                    type Error = ();

                    fn is_high(&self) -> Result<bool, Self::Error> {
                        self.is_low().map(|v| !v)
                    }

                    fn is_low(&self) -> Result<bool, Self::Error> {
                        // NOTE(unsafe) atomic read with no side effects
                        Ok(unsafe { (*$GPIOX::ptr()).idr.read().bits() & (1 << $i) == 0 })
                    }
                }
                )+
```

Indentation is wrong. I wonder why rustfmt did not complaint :thinking:

src/gpio.rs
Sh3Rm4n: Minor nitpick: I think I would move this below 

```rust
                    impl<MODE> OutputPin for $PXi<Output<MODE>> {
```

at line 668, because this is another implementation for `Output`, so this position would make more sense.

src/gpio.rs
ra-kete: I agree, we should do this in a separate PR.

But also this implementation here is not safe. The reason we pass `&mut` references for the relevant GPIO registers in is that this way we can ensure (at compile time) that we have exclusive access to them. We need this, because otherwise race conditions can happen, as explained in #37. To make this safe we'd need something else that ensures exclusive access, like a critical section.

ra-kete:
Closed in favor of #114.

PR #91, opened 2020-04-09
Clear UART overflow bit.

jonathanpallant:


jonathanpallant:
Permission to use, copy, modify, and/or distribute this bugfix for any purpose with or without fee is hereby granted.

jonathanpallant:
> I probably would have used one unsafe deref of the icr.

Yeah, good idea. FTFY.

src/serial.rs
```diff
@@ -319,21 +319,24 @@ macro_rules! hal {
                     // NOTE(unsafe) atomic read with no side effects
                     let isr = unsafe { (*$USARTX::ptr()).isr.read() };
 
+                    // NOTE(unsafe) write accessor for atomic writes with no side effects
```
strom-und-spiele, line 322: We had a discussion about using `write` in #36 .
Using it here makes sense, all I'm asking for is something like
```rust
// NOTE(unsafe, write) write accessor for atomic writes to a stateless register
```

src/serial.rs
```diff
@@ -319,21 +319,24 @@ macro_rules! hal {
...
+                    let icr = unsafe { &(*$USARTX::ptr()).icr };
+
                     Err(if isr.pe().bit_is_set() {
                         // Clear the bit by writing 1 to the clear register
-                        unsafe { (*$USARTX::ptr()).icr.write(|w| { w.pecf().set_bit() }) };
+                        icr.write(|w| { w.pecf().set_bit() });
```
strom-und-spiele, line 327: PECF offers a easy-to-understand [`.clear()` function](https://docs.rs/stm32f3xx-hal/0.4.2/stm32f3xx_hal/stm32/usart1/icr/struct.PECF_W.html)
This would (IMHO) make the comment redundant

strom-und-spiele: (This also holds for the other cases) below

PR #93, opened 2020-04-16
Add rtc module

smedellin90:
Added rtc module, with supporting helper functions in modules `rcc.rs` and `time.rs`.
It's my first type doing a PR, so apologies if the commits are a mess.


Sh3Rm4n:
Hey! :) Thank you for this PR, I'll have a look in the following days

src/rcc.rs
```diff
@@ -45,6 +46,87 @@ pub struct Rcc {
     pub cfgr: CFGR,
 }
 
+impl Rcc {
```
Sh3Rm4n, line 49: I don't know if this is the right place to add this configuration. I'm not sure about this, though. This is not really used in the rest of the `rcc` module as of now. The structure of the rcc module could take a redesign anyway, but I have a bad gut-feeling about this approach. Maybe you could explain this part a bit more, to convince me.

smedellin90:
@Sh3Rm4n ,
Much appreciate the feedback! Will be going down the list soon.

src/rtc.rs
```diff
@@ -0,0 +1,172 @@
...
+    pub fn get_date(&self) -> Date {
+        let date = self.regs.dr.read();
+        Date::new(
+            (bcd2_decode(date.yt().bits(), date.yu().bits()) + 1970).year(),
+            bcd2_decode(date.mt().bit() as u8, date.mu().bits()).month(),
+            bcd2_decode(date.dt().bits(), date.du().bits()).day(),
+        )
+    }
+
+    pub fn get_week_day(&self) -> u8 {
```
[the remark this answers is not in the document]
smedellin90: This is true. I hadn't noticed WeekDay before. I'll make the change to use WeekDay

src/rcc.rs
smedellin90: I was going back and forth with this as well. Is there a benefit to defining each register as a field within the Rcc struct? Initially, I was going to follow the current style, as well as the [stm32f1xx-hal](https://github.com/stm32-rs/stm32f1xx-hal/blob/master/src/rcc.rs) style and do something like the following:

```
impl RccExt for RCC {
    fn constrain(self) -> Rcc {
        Rcc {
            ahb: AHB { _0: () },
            apb1: APB1 { _0: () },
            apb2: APB2 { _0: () },
            cfgr: CFGR {
                hse: None,
                hclk: None,
                pclk1: None,
                pclk2: None,
                sysclk: None,
                adcclk: None,
            },
            bkp: BKP { _0: () }, // add backup domain register struct as field
            cr: CR { _0: () } // add Control register struct as field
            // etc etc
        }
    }
}
```
but because I did not fully understand why it was written that way, I went with what I considered to be more straightforward.

src/time.rs
```diff
@@ -24,6 +24,74 @@ pub struct MegaHertz(pub u32);
...
+/// Year
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Year(pub u32);
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Time {
+    pub hours: u32,
+    pub minutes: u32,
+    pub seconds: u32,
+    pub daylight_savings: bool,
```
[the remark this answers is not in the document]
smedellin90: I can see how the `daylight_savings` field can be confusing. Perhaps making a simple 'getter' and 'setter' for fmt would be a better solution?

```
/// Sets calendar clock to 24 hr format
pub fn set_fmt(&mut self) {
   self.regs.cr.modify(|_, w| w.fmt().set_bit());
}

/// Sets calendar clock to 12 hr format
pub fn clear_fmt(&mut self) {
   self.regs.cr.modify(|_, w| w.fmt().clear_bit());
}

/// Reads current hour format selection
pub fn read_fmt(&self) -> bool {
   self.regs.cr.read().fmt().bit()
}

```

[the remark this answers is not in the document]
strom-und-spiele: Going for getter/setter is not a bad idea, however I guess the confusion arose from unrelated words. When I read `daylight_savings` I'm not expecting that it has to do anything with 24/12h format. The same goes for `set_fmt`, `clear_fm` and `read_fmt`.
Why not name them `set_24h_fmt`, `set_12h_fmt` and `is_24h_fmt` ?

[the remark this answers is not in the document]
smedellin90: True, these are better names. I agree with your suggestion

src/rtc.rs
```diff
@@ -0,0 +1,172 @@
...
+
+        result.regs.cr.modify(|_, w| {
+            w
+                // sets hour format to 24 hours
+                .fmt()
+                .clear_bit()
+        });
+
+        // Prescalers set to produce a 1 hz signal
+        let (prediv_s, prediv_a) = match src {
```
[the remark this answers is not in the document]
smedellin90: These numbers came from AN4795:

temp

I like the idea of a function that takes care of choosing the correct prescalers to produce a 1 hz RTC clock source. This is possible for LSI and LSE, but the fact that HSE frequency can vary (I read that the HSE frequency can be between 4 and 32 Mhz) makes that a challenge. 

The concept of making some methods "visible" given a parameter (the RTCSrc) sounds like a good solution, if it is possible. 

For example :

if RTCSrc::LSE or RTCSrc::LSI, make the following method visible....

```
fn set_1hz_frequency{
   /// set prescalers to produce 1hz clock source
}
```

else, there is a more general function `set_frequency` that is always visible that would be used if the user is using an HSE of any frequency. The user would then supply prescalers to set appropriate frequency. 

 I think I am not using the correct words to search to see if this feature of making methods "visible" or "invisible" exists, because I am not able to find anything online. Can something like this be accomplished?

src/rcc.rs
Sh3Rm4n: This "reexports" parts of the rcc periphery, so that it can be passed to other modules and is guarded by the rust owner-ship rules instead of having to access it via unsafe. 

The A.. types resemble the peripheral clock dividers, which have to be assessed to e.g. enable the spi peripheral clock, so that it is working. A somewhat related issue could help you understand it: https://github.com/stm32-rs/stm32f3xx-hal/issues/37

src/time.rs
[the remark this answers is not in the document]
Sh3Rm4n: I like your proposed getter / setter functions. I see this as a common pattern in stm32f4xx-hal for bit values on some useful peripheral flag states.

Sh3Rm4n:
BTW I found this crate: https://crates.io/crates/rtcc at https://github.com/rust-embedded/embedded-hal/issues/207. Maybe this is useful for you, I like the idea to implement these traits for a higher level rtc access. 

Maybe it is possible to change `set_time` and `get_time` to be as raw as reasonable, and build upon these traits to give a higher functionality?

src/rtc.rs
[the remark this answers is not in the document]
Sh3Rm4n: Ah interesting to know. I would like to see a comment or make these tuples for fixed "input" frequencies (meaning LSI LSE) constants. 

I have difficulties to find the AN4795 Application note, can you provide a link?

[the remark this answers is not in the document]
smedellin90: It was a typo. I meant AN4759
Here is the link:
[https://www.st.com/resource/en/application_note/dm00226326-using-the-hardware-realtime-clock-rtc-and-the-tamper-management-unit-tamp-with-stm32-microcontrollers-stmicroelectronics.pdf](url)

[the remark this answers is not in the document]
smedellin90: Because HSE clock frequencies can vary, perhaps the user should choose a source and prescaler values to produce the desired rtc input clock speed through a configuration structure. 

eg.

```
// rtc module


/// RTC clock input source
#[derive(Clone, Copy)]
pub enum RTCSrc {
    LSE = 0b01,
    LSI = 0b10,
    HSE = 0b11,
}

/// RTC configuration struct
pub struct cfgRTC {
    src: RTCSrc,
    prediv_s: u16,
    prediv_a: u8,
}

impl cfgRTC {
    pub fn new(source: RTCSrc, prescaler_sync: u16, prescaler_async: u8) 
        -> cfgRTC 
        {
        let config  = cfgRTC {
            src: source,
            prediv_s: prescaler_sync,
            prediv_a: prescaler_async,
        };
        config
    }
}

pub struct Rtc {
    regs: RTC,
}

impl Rtc {
    pub fn rtc(regs: RTC, cfg: cfgRTC, rcc: &mut Rcc) -> Self {
        /// initialization of rtc with desired clock source and input frequency
    }
// etc etc
}

```

smedellin90:
@Sh3Rm4n , I've made changes and used the rtcc traits you previously mentioned. Is this what you had in mind? the `rtc` function within the `rtc` module is still incomplete. Still thinking about how the user should program clock speed depending on the clock source.

src/rcc.rs
smedellin90: That makes sense. I changed it so the registers needed for RTC initialization are reexported.

smedellin90:
I applied an `cfgRTC` struct similarly to what is done in [stm32l4xx-hal](https://github.com/stm32-rs/stm32l4xx-hal/blob/fad6d807b1a0a10bdf22549513d20b34ce2487fe/src/rtc.rs#L27).

David-OConnor:
What do we need to do to get this merged?

David-OConnor:
Could you provide an example? eg, in this code below, what goes in for `regs`?

```rust
    let mut rtc_ = rtc::Rtc::rtc(
        regs,
        &mut rcc.apb1,
        &mut dp.RCC.bdcr
        &mut dp.RCC.cr,
        &mut dp.RCC.csr,
        &mut dp.RCC.csr,
        rtc_cfg,
    );
```

src/rtc.rs
```diff
@@ -0,0 +1,484 @@
...
+            }
+            Hours::AM(_h) | Hours::PM(_h) => {
+                self.set_12h_fmt();
+                Ok(self.modify(|regs| regs.tr.write(|w| unsafe { w.ht().bits(ht).hu().bits(hu) })))
+            }
+        }
+    }
+
+    fn set_weekday(&mut self, weekday: u8) -> Result<(), Self::Error> {
+        if (weekday < 1) | (weekday > 7) {
```
ra-kete, line 246: You are using a bitwise OR here, but this is probably not intended, right? Logical OR is `||`

src/rtc.rs
```diff
@@ -0,0 +1,484 @@
...
+                    .yt()
+                    .bits(yt)
+                    .yu()
+                    .bits(yu)
+            });
+        }))
+    }
+
+    fn set_datetime(&mut self, date: &NaiveDateTime) -> Result<(), Self::Error> {
+        // Check if unsigned integer affects encoding to bcd
```
ra-kete, line 305: Is this still something you want to do in the PR?

src/rtc.rs
```diff
@@ -0,0 +1,484 @@
...
+        Hours::H24(h) if h > 23 => Err(Error::InvalidInputData),
+        Hours::H24(h) => Ok(bcd2_encode(h as u32)),
+        Hours::AM(h) if h < 1 || h > 12 => Err(Error::InvalidInputData),
+        Hours::AM(h) => Ok(bcd2_encode(h as u32)),
+        Hours::PM(h) if h < 1 || h > 12 => Err(Error::InvalidInputData),
+        Hours::PM(h) => Ok(bcd2_encode(h as u32)),
+    }
+}
+
+fn hours_to_u8(_hours: rtcc::Hours) -> u8 {
```
ra-kete, line 478: Why does this argument name has a leading underscore? Normally you use that for unused variables, but this on is used.

src/rtc.rs
```diff
@@ -0,0 +1,484 @@
...
+    if let rtcc::Hours::H24(h) = _hours {
+        h
+    } else {
+        panic!("_hours could not be destructured into rtc::Hours::H24(h)");
+    }
+}
```
ra-kete, line 484: That function is a bit strange. From the name I would expect it converts an `Hours` object to an `u8`. But actually it only does then the 24-hour format is used, otherwise it panics. Is there a reason why you don't handle the AM/PM cases too?

src/rtc.rs
```diff
@@ -0,0 +1,484 @@
...
+            ),
+        )
+    }
+}
+
+// Two 32-bit registers (RTC_TR and RTC_DR) contain the seconds, minutes, hours (12- or 24-hour format), day (day
+// of week), date (day of month), month, and year, expressed in binary coded decimal format
+// (BCD). The sub-seconds value is also available in binary format.
+//
+// The following helper functions are encode into BCD format from integer and
```
ra-kete, line 447: Superfluous "are"

src/rtc.rs
```diff
@@ -0,0 +1,484 @@
...
+// decode to an integer from a BCD value respectively.
+fn bcd2_encode(word: u32) -> (u8, u8) {
+    let mut value = word as u8;
+    let mut bcd_high: u8 = 0;
+    while value >= 10 {
+        bcd_high += 1;
+        value -= 10;
+    }
+    // let bcd_low = ((bcd_high << 4) | value) as u8;
```
ra-kete, line 456: Please don't leave in unused code

src/rtc.rs
```diff
@@ -0,0 +1,484 @@
...
+    }
+}
+
+// Two 32-bit registers (RTC_TR and RTC_DR) contain the seconds, minutes, hours (12- or 24-hour format), day (day
+// of week), date (day of month), month, and year, expressed in binary coded decimal format
+// (BCD). The sub-seconds value is also available in binary format.
+//
+// The following helper functions are encode into BCD format from integer and
+// decode to an integer from a BCD value respectively.
+fn bcd2_encode(word: u32) -> (u8, u8) {
```
ra-kete, line 449: So, if I understand correctly, this function converts `92` to `9` and `2`. If that's true then you could massively simplify it to:
```rust
(word / 10, word % 10)
```
But there also needs to be a (panicking) check that `word` is smaller than 100.

src/rtc.rs
```diff
@@ -0,0 +1,484 @@
...
+    while value >= 10 {
+        bcd_high += 1;
+        value -= 10;
+    }
+    // let bcd_low = ((bcd_high << 4) | value) as u8;
+    let bcd_low = value as u8;
+    (bcd_high, bcd_low)
+}
+
+fn bcd2_decode(fst: u8, snd: u8) -> u32 {
```
ra-kete, line 461: Similarly, this can be simplified to:
```rust
fst * 10 + snd
```

src/rtc.rs
```diff
@@ -0,0 +1,484 @@
...
+}
+
+fn hours_to_register(hours: Hours) -> Result<(u8, u8), Error> {
+    match hours {
+        Hours::H24(h) if h > 23 => Err(Error::InvalidInputData),
+        Hours::H24(h) => Ok(bcd2_encode(h as u32)),
+        Hours::AM(h) if h < 1 || h > 12 => Err(Error::InvalidInputData),
+        Hours::AM(h) => Ok(bcd2_encode(h as u32)),
+        Hours::PM(h) if h < 1 || h > 12 => Err(Error::InvalidInputData),
+        Hours::PM(h) => Ok(bcd2_encode(h as u32)),
```
ra-kete, line 474: This would be more readable without the if-guards:
```rust
   match hours {
        Hours::H24(h @ 0..=23) => Ok(bcd2_encode(h as u32)),
        Hours::AM(h @ 1..=12) | Hours::AM(h @ 1..=12) => Ok(bcd2_encode(h as u32)),
        _ => Err(Error::InvalidInputData),
}
```

src/rcc.rs
```diff
@@ -159,7 +168,20 @@ mod usb_clocking {
...
+
+impl BDCR {
+    pub(crate) fn bdcr(&mut self) -> &rcc::BDCR {
+        // NOTE(unsafe) this proxy grants exclusive access to this register
+        unsafe { &(*RCC::ptr()).bdcr }
+    }
+}
+
 /// Clock configuration
+#[derive(Clone, Copy)]
```
ra-kete, line 184: We must not allow cloning of the `CFGR` struct. This struct is responsible for making sure that the clocks are only configured once. When you call `freeze()` it consumes the struct and thereby ensures the clocks won't change anymore, so other code can rely on that. If you make `CFGR` clonable you could suddenly have multiple instances of this struct around, which would allow you to change the clocks and break the assumption that they are stable

src/rcc.rs
```diff
@@ -591,6 +613,30 @@ impl CFGR {
...
 }
 
+/// RCC Clock Control register (RCC_CR)
+pub struct CR {
+    _0: (),
+}
+
+impl CR {
+    pub(crate) fn cr(&mut self) -> &rcc::CR {
+        // NOTE(unsafe) this proxy grants exclusive access to this register
```
ra-kete, line 623: That's not true. If you look at `CFGR`'s `freeze()` method, that one also modifies the `CR` register, so the access here is not exclusive at all. I guess `CFGR` already assumes exclusive access to `CR` though that's not documented unfortunately. Which means this `CR` proxy shouldn't exist. Or at least it shouldn't exist as long as `freeze()` was not called

src/rtc.rs
```diff
@@ -0,0 +1,484 @@
+/*!  Interface to the real time clock */
```
ra-kete, line 1: Please use the single-line comment style for all comments, i.e. `//! ...` in this case

src/rtc.rs
```diff
@@ -0,0 +1,484 @@
...
+
+use crate::rcc::{APB1, BDCR, CR, CSR};
+use crate::stm32::{PWR, RTC};
+use rtcc::{Datelike, Hours, NaiveDate, NaiveDateTime, NaiveTime, Rtcc, Timelike};
+
+/// Invalid input error
+// #[derive(Clone, Copy)]
```
ra-kete, line 8: Did you forget to remove this?

src/rtc.rs
```diff
@@ -0,0 +1,484 @@
...
+#[derive(Debug)]
+pub enum Error {
+    InvalidInputData,
+}
+
+/// RTC clock input source
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum RTCSrc {
```
ra-kete, line 16: Check out this RFC for naming conventions in Rust: https://github.com/rust-lang/rfcs/blob/master/text/0430-finalizing-naming-conventions.md

Notably acronyms count as one word for the CamelCase notation, so did should be `RtcSrc`. But there is no need to obfuscate the name just to save three characters here, so `RtcSource` would be what I prefer.

src/rtc.rs
```diff
@@ -0,0 +1,484 @@
...
+    LSE = 0b01,
+    LSI = 0b10,
+    HSE = 0b11,
```
ra-kete, line 19: I don't think to use these bit patterns anywhere? If so, I would just leave them out.

src/rtc.rs
```diff
@@ -0,0 +1,484 @@
...
+}
+
+/// RTC configuration struct
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub struct cfgRTC {
```
ra-kete, line 24: This name also doesn't follow the naming conventions. How about `RtcConfig`?

src/rtc.rs
```diff
@@ -0,0 +1,484 @@
...
+}
+
+/// LSI clock source with prescalers
+/// set clock frequency to 1 hz
+impl Default for cfgRTC {
+    fn default() -> Self {
+        cfgRTC {
+            src: RTCSrc::LSI,
+            prediv_s: 311,
+            prediv_a: 127,
```
ra-kete, line 37: You discussed this with @Sh3Rm4n before. Like him I think there should be a reference to the relevant manual and section you got these magic numbers from documented.

ra-kete:
> Thoughts on merging this now? We can clean up details later if they arise.

I would rather have a clean initial implementation, since anything else will fall back on us by making this crate harder to maintain. Hoping that someone will clean up later doesn't work out in my experience, because everyone prefers to work on new features instead ;)

If something should be merged faster, then I think the way to go is by merging a (good) MVP first and adding more features later.

David-OConnor:
I'm working off a variant of this and tweaking as I go. I'll let you know what I come up with. Excited to get this cleaned up and merged.

David-OConnor:
Do you think RTC wakeup code [like this](https://github.com/David-OConnor/stm32f3xx-util/blob/master/src/interrupt_.rs#L471) should go in this RTC module, or in an interrupt module?

src/rtc.rs
```diff
@@ -0,0 +1,484 @@
...
+                    .bits(cfg_rtc.prediv_s)
+                    .prediv_a()
+                    .bits(cfg_rtc.prediv_a)
+            });
+        });
+        result
+    }
+
+    fn enable_lsi(csr: &mut CSR) {
+        csr.csr().write(|w| w.lsion().set_bit());
```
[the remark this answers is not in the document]
David-OConnor: My limited knowledge points to the same. The calls to `write` followed by unsafe blocks are probably correct, but the others should be modify. (?)

ra-kete:
I can't say much to RTC specifics because I've not read up to that yet.

What I can say is that usually how it works is that a HAL crate provides abstractions over peripherals. When you create an instance of such an abstraction, you have to give it ownership over the peripheral registers. The abstraction basically wraps these registers and provides a safe API for them. Peripheral abstractions interact with one another, but they don't access the registers of another peripheral directly, only the safe methods.

So in this case, the relevant peripheral abstraction would be `Rtc`. When instantiating it, you have to give it the `RTC` register, at which point `Rtc` has exclusive access over it. `Rtc` is responsible for managing that peripheral now. It won't give out the `RTC` register for other code to modify, since that would make it impossible (or at least very hard) to provide a safe API. But that's okay, since other components shouldn't have a need to access the register directly, they can always call methods on the `Rtc` instead.

So no, I don't think we should provide public access to `RTC`. What we could provide is an `Rtc::free` method that consumes the `Rtc` and gives its owned registers back in the process. I think this is good practice to have for all peripheral abstractions (see https://docs.rust-embedded.org/book/design-patterns/hal/interoperability.html#c-free) but I don't think we're doing that consistently yet.

As a final note, it is always possible to get a register using unsafe code, e.g.: `let rtc = unsafe { &*RTC::ptr() };`. So in a pinch you can add functionality missing from the HAL that way. Of course, the long-term solution should be adding the missing functionality to the HAL.

David-OConnor:
That's a bit of a 2-for-1 response: The interrupt (wakeup, alarm, tamper etc) functionality needs to be baked into this module.

David-OConnor:
Btw, I'm glad this uses the `rtcc` module. It was a recent addition to the HAL ecosystem that fits the role of abstracting RTC to any architecture. I think it should be officially in `embedded-hal`.

Thoughts on adding the `setup_rtc_wakeup` fn in my code above directly to this PR, since it uses the RTC periph this owns?

Sh3Rm4n:
> Btw, I'm glad this uses the rtcc module. It was a recent addition to the HAL ecosystem that fits the role of abstracting RTC to any architecture. I think it should be officially in embedded-hal.

Off-topic: Regarding `rtcc` as a potential candidate for the embedded-hal. I wonder if we should feature gate these crates until we official support in the `embedded-hal` to avoid breaking changes or should we just go on and if we switch the implmentation back to embedded-hal, if the trait is upstreamed and

eldruin:
Hi, nice to see `rtcc` being used for an MCU HAL implementation as well :). Please let me know if it fits well and if there are things that could be improved.

I do not want to hold it back but I do not think it will be merged into `embedded-hal` any time soon if at all so I would not hold my breath on that. I think it would be interesting to see adoption among HAL implementations for several MCUs grow first to iron out any difficulties and see of how it fits.

Anyway feel free to express your opinion at https://github.com/rust-embedded/embedded-hal/issues/207 as well.

src/rtc.rs
```diff
@@ -0,0 +1,484 @@
...
+        apb1: &mut APB1,
+        bdcr: &mut BDCR,
+        cr: &mut CR,
+        csr: &mut CSR,
+        cfg_rtc: cfgRTC,
+    ) -> Self {
+        match &cfg_rtc.src {
+            RTCSrc::LSI => Rtc::enable_lsi(csr),
+            RTCSrc::HSE => Rtc::enable_hse(cr, false),
+            RTCSrc::LSE => Rtc::enable_lse(bdcr, false),
```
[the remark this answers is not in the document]
smedellin90: That is correct, the reason for exposing `CSR` and `CR` is to allow clock source selection. The LSE is not included on the STM32F303 Discovery board, so I initially wanted to make the rtc module flexible and allow the user to choose. That being said, if I had to choose one for support, it would be the LSE. I agree with you.

src/rtc.rs
smedellin90: Will not be a problem with only LSE support

src/rtc.rs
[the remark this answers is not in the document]
smedellin90: Thank you for catching this. I'll look through the code to see if I confused it anywhere else.

smedellin90:
> Do you think RTC wakeup code [like this](https://github.com/David-OConnor/stm32f3xx-util/blob/master/src/interrupt_.rs#L471) should go in this RTC module, or in an interrupt module? Note that the `artc` var there is `dp.RTC`, ie what this module calls the `regs` field.
> 
> It's not working yet, but I think I'm almost there, since I've done everything I can find in the ref man and rtc man.
> 
> Thoughts on making `reg` public? ie using the `Rtc` struct then blocks you from (mutably borrowing) the `RTC` peripheral elsewhere.

@David-OConnor , I believe it makes the most sense to apply the RTC wakeup code in a separate EXTI module, as it is done [here](https://github.com/stm32-rs/stm32l0xx-hal/blob/master/src/exti.rs) .

David-OConnor:
Then we'd need to make ref public.

ra-kete:
Closing in favor of #136.

PR #94, opened 2020-04-17
Deal with a few warnings

Sh3Rm4n:


Sh3Rm4n:
Thanks for the investigation. I just followed clippy's suggestions but to see the difference in practice is quite interesting. 

I do not know much about the specifics of the relation between `clock cycle` to `instruction` but my assumption is, that at least on the arm architecture that this is more or less a 1:1 relation. I stand to be corrected, though.

PR #96, opened 2020-04-17
Crate a ci job for every chip feature

Sh3Rm4n:


strom-und-spiele:
I fail to see the improvement; while `ci/build.py` pulled all chipfeatures from `Cargo.toml` this list needs manual maintanance.

Sh3Rm4n:
Yeah, that's true. I just tried it out to see if we have build time improvements. But there are none, as we can see. The current version is better as you said, because we won't have to update the travis file manually :)

ra-kete:
There has been a lot of talk about GitHub Actions lately. I have no experience with that, but everyone says it's so much better than Travis, so maybe we should give it a try.

Sh3Rm4n:
I've heard about that, and I am excited about GitHub Actions. I've run into some issues with travis, so I would like to move away from it.   (I still don't know why two instances of travis are spawned running the same CI twice)

My plan was to update this PR to GitHub Actions, but I have not yet found the time to try it out.

ra-kete:
So should we close this PR and create an issue for migrating to GA instead?

Sh3Rm4n:
Yeah, that sounds reasonable :)

ra-kete:
Closing in favor of #113.

PR #97, opened 2020-04-26
Update stm32f3 to v0.11.0

Sh3Rm4n:
Remove now unneeded unsafe from the flash state configuration in `src/rcc.rs` and use `HCLK` instead of `SYSCLK` as the reference manual suggests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ cortex-m = "0.6"
 cortex-m-rt = "0.6"
 embedded-hal = "0.2"
 nb = "0.1"
-stm32f3 = "0.10"
+stm32f3 = "0.11"
 
 [dependencies.bare-metal]
 version = "0.2"
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -524,18 +524,17 @@ impl CFGR {
 
         assert!(pclk2 <= 72_000_000);
 
-        // adjust flash wait states
-        unsafe {
-            acr.acr().modify(|_, w| {
-                w.latency().bits(if sysclk <= 24_000_000 {
-                    0b000
-                } else if sysclk <= 48_000_000 {
-                    0b001
-                } else {
-                    0b010
-                })
-            })
-        }
+        // Adjust flash wait states according to the
+        // HCLK frequency (cpu core clock)
+        acr.acr().modify(|_, w| {
+            if hclk <= 24_000_000 {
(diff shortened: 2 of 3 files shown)


src/rcc.rs
```diff
@@ -338,17 +338,15 @@ impl CFGR {
...
-                w.latency().bits(if sysclk <= 24_000_000 {
-                    0b000
-                } else if sysclk <= 48_000_000 {
-                    0b001
-                } else {
-                    0b010
-                })
-            })
-        }
+        acr.acr().modify(|_, w| {
```
strom-und-spiele, line 341: All those magic numbers make me nervous.
I'm not sure if it helps that they are [documented but not explained in stm32f3](https://docs.rs/stm32f3/0.11.0/src/stm32f3/stm32f303/flash/acr.rs.html#16-23)

src/rcc.rs
```diff
@@ -338,17 +338,15 @@ impl CFGR {
...
+            if hclk <= 24_000_000 {
+                w.latency().ws0()
+            } else if hclk <= 48_000_000 {
+                w.latency().ws1()
```
strom-und-spiele, line 345: All those magic numbers make me nervous.
I'm not sure if it helps that they are [documented but not explained in stm32f3](https://docs.rs/stm32f3/0.11.0/src/stm32f3/stm32f303/flash/acr.rs.html#16-23)

Sh3Rm4n: Yeah, the documentation of the stm32f3 is always very brief. But more detailed explanation is not the intent of the `stm32f3` crate, as this is auto-generated code from svd files. A more detailed explanation is always in the reference manual. However, it does not hurt to add documentation here, to explain, why this is needed.

Cargo.toml
```diff
@@ -23,7 +23,7 @@ cortex-m = "0.6"
 cortex-m-rt = "0.6"
 embedded-hal = "0.2"
 nb = "0.1"
-stm32f3 = "0.10"
+stm32f3 = "0.11.0"
```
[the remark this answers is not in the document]
Sh3Rm4n: For consistency, I should use "0.11". Thanks for pointing this out :) 

Cargo's version always translates to the "^0.11.0" [semver](https://semver.org/) version number, unless stated otherwise. So it should not make a difference, as `0.11.1` would automatically be pulled in, when such a version would be released.

PR #98, opened 2020-05-01
Add reclock functionality for SPI

ceigel:
Some devices (eg SD cards) require the initialization to be done a at a lower frequency and allow the frequency to be later increased. Thus a re-clock functionality has been added. The tm4c123x-hal also does this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+- SPI support for reclock after initialization ([#98](https://github.com/stm32-rs/stm32f3xx-hal/pull/98))
+
 ## [v0.5.0] - 2020-07-21
 
 ### Added
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -5,6 +5,7 @@ use core::ptr;
 use crate::hal::spi::FullDuplex;
 pub use crate::hal::spi::{Mode, Phase, Polarity};
 use crate::pac::{SPI1, SPI2, SPI3};
+use crate::stm32::spi1;
 
 use crate::gpio::gpioa::{PA5, PA6, PA7};
 #[cfg(any(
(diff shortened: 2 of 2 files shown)


src/spi.rs
```diff
@@ -114,6 +114,20 @@ pub struct Spi<SPI, PINS> {
     pins: PINS,
 }
 
+fn compute_clock_variant(clocks: Hertz, freq: Hertz) -> u8 {
```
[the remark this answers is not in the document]
Sh3Rm4n: Also rename to 
```suggestion
fn compute_baud_rate(clocks: Hertz, freq: Hertz) -> u8 {
```

src/spi.rs
```diff
@@ -200,6 +204,17 @@ macro_rules! hal {
                 pub fn free(self) -> ($SPIX, (SCK, MISO, MOSI)) {
                     (self.spi, self.pins)
                 }
+
+                pub fn reclock<F>(&mut self, freq: F, clocks: Clocks)
+                    where F: Into<Hertz>
+                {
+                    self.spi.cr1.modify(|_, w| w.spe().disabled());
```
Sh3Rm4n, line 211: Are there any side effects changing the frequency, or is it sufficient to disable and enable it right after it? Any delay needed?

src/spi.rs
```diff
@@ -200,6 +204,17 @@ macro_rules! hal {
                 pub fn free(self) -> ($SPIX, (SCK, MISO, MOSI)) {
                     (self.spi, self.pins)
                 }
+
+                pub fn reclock<F>(&mut self, freq: F, clocks: Clocks)
```
Sh3Rm4n, line 208: Could use a doc-comment.

src/spi.rs
ceigel: Unfortunately I didn't find any documentation about this. 
All I can say is that I tested this with an SD card using the embedded-sdmmc library. I initialize the card at 400KHz, then I switch to 16Mhz and data is red without a problem. I'm reading mp3s from the SD-card and playing them (after decoding) at the same time. Would there be a problem I think I'd notice it. If I leave the card at 400Khz the playing doesn't go well. I tried this also by initializing the card directly at 8 Mhz, and I would get random things and strange crashes in the sdmmc library. I also believe this depends on the device at the other end. With SD cards this works, with other devices it might not.

ceigel:
@Sh3Rm4n Thank you for your review. I believe I addressed all your comments.

ceigel:
@Sh3Rm4n thank you for your review. I don't have the rights to merge the PR. Could you do it for me?

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Support for safe one-shot DMA transfers ([#86](https://github.com/stm32-rs/stm32f3xx-hal/pull/86))
 - DMA support for serial reception and transmission ([#86](https://github.com/stm32-rs/stm32f3xx-hal/pull/86))
 - ADC support for `stm32f303` devices ([#47](https://github.com/stm32-rs/stm32f3xx-hal/pull/47))
+- SPI support for reclock after initialization ([#98](https://github.com/stm32-rs/stm32f3xx-hal/pull/98))
```
Sh3Rm4n, line 18: This is the wrong place, as 0.5.0 is already released. You'd have to move it under `Unreleased` :)

PR #100, opened 2020-05-09
CAN Support

zklapow:
This PR is fairly incomplete still, though it is functional.

I created a (temp) create called `embedded-hal-can` to try implementing the traits currently in https://github.com/rust-embedded/embedded-hal/pull/77 though that PR is so old now, I am not sure its going to get much traction (and I am not totally sure I love the structure of those traits anyhow) so I may remove that piece.

I have been testing this on an `STM32f302C8T6`, I am working on simplifying my "blinky with can" into an example of how to use this.

todos:
- [x] Generify this to support other stm32f3xx devices

future improvements:
- Some sort of baud rate param heuristics (maybe just embed a table for standard rates?)
- Improved filter handling (the current implementation "wastes" filter registers)

Wanted to post this now though as I have been sitting on it for a while and I figured it would be good to at least get thoughts/feedback.


src/can.rs
```diff
@@ -0,0 +1,511 @@
...
+
+use crate::gpio::gpioa;
+use crate::gpio::AF9;
+use crate::rcc::APB1;
+use crate::stm32;
+use nb;
+use nb::Error;
+
+use core::sync::atomic::{AtomicU8, Ordering};
+use stm32f3::stm32f302::CAN;
```
Sh3Rm4n, line 11: Has to be adjusted to use the controller declared by the feature flag.

src/can.rs
```diff
@@ -0,0 +1,511 @@
...
+const EXID_MASK: u32 = 0b11111111111000000000000000000;
+const MAX_EXTENDED_ID: u32 = 0x1FFFFFFF;
+
+#[derive(Debug, Copy, Clone, Hash, Eq, PartialEq)]
+pub enum CanId {
+    BaseId(u16),
+    ExtendedId(u32),
+}
+
+#[derive(Debug, Clone, Hash, Eq, PartialEq)]
```
Sh3Rm4n, line 22: Is `#derive(Hash)` needed for anything?

Cargo.toml
```diff
@@ -24,6 +24,8 @@ cortex-m-rt = "0.6"
 embedded-hal = "0.2"
 nb = "0.1"
 stm32f3 = "0.10"
+typenum = "1.11.2"
```
Sh3Rm4n, line 27: I don't see were `typenum` is used.

Cargo.toml
```diff
@@ -24,6 +24,8 @@ cortex-m-rt = "0.6"
...
+embedded-hal-can = "0.1.0"
```
Sh3Rm4n, line 28: Hm. As this is an experimental crate until this is added to the `embedded-hal` we should feature gate the CAN implementation as a whole?

But I'm happy to include it into this hal, so that more people can experiment with the can abstraction! :)

src/can.rs
```diff
@@ -0,0 +1,511 @@
...
+            CanFilterData::ExtendedMaskFilter(_, mask) => Some(mask << 3),
+            CanFilterData::MaskFilter(_, _mask) => None, // TODO: We should be able to fill this register with a second filter/mask pair
+            CanFilterData::IdFilter(_id) => None, // TODO: This sucks, we need more info here to figure out the correct value of fr2
+        }
+    }
+}
+
+impl Can {
+    pub fn new(
+        can: stm32::CAN,
```
Sh3Rm4n, line 192: Even though rust has the `new` convention for constructors, naming the constructor as the peripheral name is really common in the rust hal implementations. This is import e.g. for SPI to differentiate between `spi1, spi2, ...`. As stm32f3 devices do not have multiple can peripherals this is not as important, but maybe we should stick to it and name it `can`?

src/can.rs
```diff
@@ -0,0 +1,511 @@
...
+        rx: gpioa::PA11<AF9>,
+        tx: gpioa::PA12<AF9>,
+        apb1: &mut APB1,
+    ) -> Can {
+        apb1.enr().modify(|_, w| w.canen().set_bit());
```
Sh3Rm4n, line 197: The can peripheral is not well documented atm as of [stm32-rs](https://stm32-rs.github.io/stm32-rs/STM32F302.html#CAN), but this could be named

```suggestion
        apb1.enr().modify(|_, w| w.canen().enabled());
```

[doc link](https://docs.rs/stm32f3/0.11.0/stm32f3/stm32f303/rcc/apb1enr/struct.CANEN_W.html#method.enabled)

Maybe you could try to apply these to methods, where it is possible right now :)

src/can.rs
```diff
@@ -0,0 +1,511 @@
...
+    type Filter = CanFilter;
+}
+
+impl embedded_hal_can::Transmitter for CanTransmitter {
+    fn transmit(
+        &mut self,
+        frame: &Self::Frame,
+    ) -> Result<Option<Self::Frame>, nb::Error<Self::Error>> {
+        if let Some(_) = frame.data() {
+            assert!(
```
Sh3Rm4n, line 299: Hm. I think  we should generally thrive towards not using asserts inside this hal. (There are enough other places in the hal, where we do use asserts) But  this assert is unfortunate, as we need the `nb::Error` to return. 

Do you have any ideas?

src/can.rs
```diff
@@ -0,0 +1,511 @@
...
+            match frame.id {
+                CanId::BaseId(id) => tx.tir.modify(|_, w| unsafe { w.stid().bits(id) }),
+                CanId::ExtendedId(id) => tx.tir.modify(|_, w| unsafe {
+                    let hb = ((id & EXID_MASK) >> 18) as u16;
+                    w.exid().bits(id).stid().bits(hb).ide().set_bit()
+                }),
+            }
+
+            match frame.data() {
+                Some(_) => unsafe {
```
Sh3Rm4n, line 330: `if let Some(..)` could be used here, I guess :thinking:

src/can.rs
```diff
@@ -0,0 +1,511 @@
...
+                    tx.tir.modify(|_, w| w.rtr().clear_bit());
+                },
+                None => tx.tir.modify(|_, w| w.rtr().set_bit()),
+            }
+
+            // Request the frame be sent!
+            tx.tir.modify(|_, w| w.txrq().set_bit());
+
+            // TODO: we can prob not wait for txok?
+            while match i {
```
Sh3Rm4n, line 357: I lost the meaning of `i` here. Maybe a more descriptive name is needed.

src/can.rs
```diff
@@ -0,0 +1,511 @@
...
+        can.fmr.modify(|_, w| w.finit().clear_bit());
+    }
+
+    fn clear_filter(&mut self) {
+        self.set_filter(CanFilter::new(CanFilterData::AcceptAll))
+    }
+}
+
+impl CanFrame {
+    pub fn new_with_len(id: CanId, data: &[u8], length: usize) -> CanFrame {
```
Sh3Rm4n, line 487: Just curious. Why must the data slice be converted into an array? Slice has a [`len()`](https://doc.rust-lang.org/std/primitive.slice.html#method.len) method. Maybe you could give me some insight :)

src/can.rs
Sh3Rm4n: While reading through the code, isn't this assert already handled in `CanFrame`?

src/can.rs
zklapow: I should likely limit the visibility of this method and it could be simplified a bit... The reason this exists is mainly to do with lifetimes and there are sort of two reasons for it:

1. is for reading, we allocate an array always of length 8 bytes (the max CAN frame size), because we cant do dynamic allocation and then use a subslice of it, but now that I think more about it, we can likely just consume that array in this case and avoid the reference.
2. Is because of how we need to implement `embedded_hal_can::Frame`. Having every frame have a 8 byte backing array was just the simplest way to handle this without getting into using `generic_array` or something, but ultimately I may do that, because it would make the code clearer.

src/can.rs
zklapow: ya ill remove this

src/can.rs
zklapow: 👍 a bunch of this was written before some upgrades to the svd-to-rust stuff which I suspect is the source of this.

Cargo.toml
zklapow: ya my plan was to put this whole thing behind a feature flag once it was more merge-ready, but intellij doesn't totally know how to handle feature flags ATM so I have avoided that for ease of development.

zklapow:
Added a simple "blinky" example, heres a GIF of it in action: 

Image

zklapow:
@Sh3Rm4n this should be ready for another look when you have some time, it should be mergeable in its current state.

A couple notes:

- I updated to use `ArrayVec` to give the `CanFrame` a slightly simpler API (no need to remember that there is an extra `length` field there!). I tried to use `generic_array` but I don't think that will be possible without having `transmit` be generic of the frame size. Given that a frame is at most 8 bytes, having to allocate the full backing array seemed OK. (I also considered introducing some sort of allocator dedicated to canframes but.... that seemed extreme)
- Everything should be behind the `can` feature flag now, so it is disabled by default.

zklapow:
@Sh3Rm4n any thoughts on another review or merging this?

zklapow:
I moved those around a bit to be more clear, I do think they would be interesting to tackle in the future but are sort of hard and don't really affect the functionality here much. This should be useable and mergeable as-is.

Cargo.toml
```diff
@@ -25,6 +25,15 @@ embedded-hal = "0.2"
...
 stm32f3 = "0.10"
 
+[dependencies.embedded-hal-can]
+version = "0.1.0"
+optional = true
+
+[dependencies.arrayvec]
+version = "0.5.1"
+default-features = false
+optional = true
```
Sh3Rm4n, line 35: [heapless](https://crates.io/crates/heapless) is an alternative and more widely used in the embedded community, FYI. 

But I do not care and haven't used both enough to have an opion, if one or the other would be better for this use case.

Cargo.toml
```diff
@@ -93,7 +105,7 @@ required-features = ["stm32f303xc"]
 
 [[example]]
 name = "toggle"
-required-features = ["rt", "stm32f303xc"]
+required-features = ["rt", "stm32f302xc"]
```
Sh3Rm4n, line 108: Should be reverted I guess :)

Cargo.toml
```diff
@@ -102,3 +114,7 @@ required-features = ["rt", "stm32f303xc", "stm32-usbd"]
 [[example]]
 name = "spi"
 required-features = ["stm32f303"]
+
+[[example]]
+name = "can"
+required-features = ["rt", "can", "device-selected"]
```
Sh3Rm4n, line 120: As the [example](https://github.com/stm32-rs/stm32f3xx-hal/pull/100/files#diff-f005bfff7df113692409278f2a71eec9R7) explicitly uses the `stm32f303` we should requre the "stm32f303" feature instead of the "device-sected" feature: 

```suggestion
required-features = ["rt", "can", "stm32f303"]
```

"device-seletected" feature is not really meant to be used outside the crate and is an implementation detail.

examples/can.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    let mut led0 = gpiob
+        .pb15
+        .into_push_pull_output(&mut gpiob.moder, &mut gpiob.otyper);
+    led0.set_high().unwrap();
+
+    let filter = CanFilter::from_mask(0b100, ID as u32);
+    rx0.set_filter(filter);
+
+    // Watchdog makes sure this gets restarted periodically if nothing happens
```
Sh3Rm4n, line 60: Nice, this is also now an example for the watchdog :+1:

examples/can.rs
```diff
@@ -0,0 +1,93 @@
...
+        .cfgr
+        .use_hse(32.mhz())
+        .sysclk(32.mhz())
+        .pclk1(16.mhz())
+        .pclk2(16.mhz())
+        .freeze(&mut flash.acr);
+
+    // Configure CAN RX and TX pins (AF7)
+    let can_rx = gpioa.pa11.into_af9(&mut gpioa.moder, &mut gpioa.afrh);
+    let can_tx = gpioa.pa12.into_af9(&mut gpioa.moder, &mut gpioa.afrh);
```
Sh3Rm4n, line 42: AF7 or AF9?

src/can.rs
```diff
@@ -0,0 +1,498 @@
+pub use embedded_hal_can::{self, Filter, Frame, Id, Receiver, Transmitter};
+
+use crate::gpio::gpioa;
+use crate::gpio::AF9;
+use crate::rcc::APB1;
+use crate::stm32;
+use arrayvec::ArrayVec;
+use nb;
+use nb::Error;
```
Sh3Rm4n, line 9: Could also be written as

```suggestion
use nb::{self, Error};
```

src/can.rs
```diff
@@ -0,0 +1,498 @@
...
+
+use core::sync::atomic::{AtomicU8, Ordering};
+
+const EXID_MASK: u32 = 0b11111111111000000000000000000;
+const MAX_EXTENDED_ID: u32 = 0x1FFFFFFF;
```
Sh3Rm4n, line 14: Would be more readable this way:

```suggestion
const EXID_MASK: u32 = 0b11111_11111100_00000000_00000000;
const MAX_EXTENDED_ID: u32 = 0x1FFF_FFFF;
```

src/can.rs
```diff
@@ -0,0 +1,498 @@
...
+pub struct CanFilter {
+    data: CanFilterData,
+    index: Option<u8>,
+}
+
+static FILTER_INDEX: AtomicU8 = AtomicU8::new(0);
+
+pub struct Can {
+    can: stm32::CAN,
+    _rx: gpioa::PA11<AF9>,
```
Sh3Rm4n, line 51: What the meaning / purpose of these underscores?

src/can.rs
```diff
@@ -0,0 +1,498 @@
...
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub struct CanFilter {
+    data: CanFilterData,
+    index: Option<u8>,
+}
+
+static FILTER_INDEX: AtomicU8 = AtomicU8::new(0);
+
+pub struct Can {
```
Sh3Rm4n, line 49: I would like to see more comments / documentation, especially on the public interfaces, like this struct :)

src/can.rs
```diff
@@ -0,0 +1,498 @@
...
+            match frame.id {
+                CanId::BaseId(id) => tx.tir.modify(|_, w| unsafe { w.stid().bits(id) }),
+                CanId::ExtendedId(id) => tx.tir.modify(|_, w| unsafe {
+                    let hb = ((id & EXID_MASK) >> 18) as u16;
+                    w.exid().bits(id).stid().bits(hb).ide().set_bit()
+                }),
+            }
+
+            if let Some(_) = frame.data() {
+                unsafe {
```
Sh3Rm4n, line 322: A comment on this `unsafe` instance would be great. And couldn't this `unsafe` scope be limited insid3 the `math j { ... }`?

src/can.rs
```diff
@@ -0,0 +1,498 @@
...
+            return Ok(None);
+        }
+
+        Err(nb::Error::WouldBlock)
+    }
+}
+
+impl Receiver for CanFifo {
+    fn receive(&mut self) -> Result<Self::Frame, Error<Self::Error>> {
+        let can = can_register();
```
Sh3Rm4n, line 358: I would make this explicit and not hide it behind a function, as it is done everywhere else in this hal IIRC

```suggestion
        let can = unsafe { &*stm32::CAN::ptr() };
```

src/can.rs
zklapow: I'll add a note, there are actually 2 unsafe calls here which is the reason for the outer scope (the tdtr call below also calls the unsafe `bits`) but these can be split. The note here will probably be somewhat silly, I'd have to do some more digging but I wonder if we could get svd-to-rust to generate safe calls for these registers, as the only actual way to update them is via a full register write

src/can.rs
zklapow: this just indicates that these fields are unused

zklapow:
I have updated this to address the open comments. I added a bunch of notes about the usages of unsafe but a lot of those notes are semi silly but calling `bits()` is just always unsafe but also the only way to access those registers.

David-OConnor:
Merge!

zklapow:
@David-OConnor would love to tho I dont have access 👍 

@Sh3Rm4n or @ra-kete I have merged master to address some conflicts as well as updated the changelog, and the build is passing. Any chance one of you could merge this for me when you have a moment? I think all of the comments should be addressed.

ra-kete:
@Sh3Rm4n has reviewed this, so if he gives his thumbs-up for merging I'm fine too.

src/can.rs
```diff
@@ -0,0 +1,510 @@
...
+        let transmitter = CanTransmitter {
+            _can: self.can,
+            _rx: self._rx,
+            _tx: self._tx,
+        };
+
+        return (transmitter, fifo0, fifo1);
+    }
+
+    pub fn release(self) -> stm32::CAN {
```
Sh3Rm4n, line 286:
```suggestion
    pub fn free(self) -> stm32::CAN {
```

And it should release `gpioa::PA11<AF9>` and    `gpioa::PA12<AF9>` as well.

Sh3Rm4n (approved):
Hey, sorry for the late response. 

Minor nits and [this](https://github.com/stm32-rs/stm32f3xx-hal/pull/100/files#r456956829) is not addressed, yet. 

But overall LGTM. :)

zklapow:
updated, the docs are still a bit sparse but unfortunately I don't have a ton of time to work on this ATM.

Sh3Rm4n:
Nice. Thanks for the contribution. Finally i've time to merge this 🎉 

Thanks for your patience. Further things can still be improved later on, if necessary.

PR #105, opened 2020-06-02
Implemented InputPin trait for output pins

David-OConnor:
This enables things like OneWire. Equiv of [this PR](https://github.com/stm32-rs/stm32f4xx-hal/pull/18) from the 4xx lib.

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -398,6 +398,20 @@ macro_rules! gpio {
                 }
 
                 #[cfg(feature = "unproven")]
+                impl<MODE> InputPin for $PXx<Output<MODE>> {
+                    type Error = Infallible;
+
+                    fn is_high(&self) -> Result<bool, Self::Error> {
+                        Ok(!self.is_low()?)
+                    }
+
+                     fn is_low(&self) -> Result<bool, Self::Error> {
+                        // NOTE(unsafe) atomic read with no side effects
+                        Ok(unsafe { (*$GPIOX::ptr()).idr.read().bits() & (1 << self.i) == 0 })
+                    }
+                }
+
+                #[cfg(feature = "unproven")]
                 impl<MODE> InputPin for $PXx<Input<MODE>> {
                     type Error = Infallible;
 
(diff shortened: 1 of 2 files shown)


jeru:
+1. I also need this functionality, to implement a weird "bidirectional" SPI for ST7735 TFT.

And it is already implemented in the "brother" crate stm32f4xx-hal (https://github.com/stm32-rs/stm32f4xx-hal/blob/b6f010538a4f86db207b23440d3053339ce08397/src/gpio.rs#L179).

David-OConnor:
Status?

Sh3Rm4n:
Sorry for the long delay. I'll hope to catch up again in the next days :) 

What about this concern: https://github.com/stm32-rs/stm32f3xx-hal/issues/84#issuecomment-646654451?

I havn't thought it through myself. Implementing for `OpenDrain` seems fine though. Maybe we could mark the `PushPull` implementation `unsafe` or at least add a comment to use this function with caution.

src/rcc.rs
```diff
@@ -66,7 +66,7 @@ pub struct APB1 {
 }
 
 impl APB1 {
-    pub(crate) fn enr(&mut self) -> &rcc::APB1ENR {
```
Sh3Rm4n, line 69: Why is this changed? :)

src/gpio.rs
```diff
@@ -397,6 +397,20 @@ macro_rules! gpio {
                     }
                 }
 
+                #[cfg(feature = "unproven")]
+                impl<MODE> InputPin for $PXx<Output<MODE>> {
```
Sh3Rm4n, line 401: `impl` is missing for `PXx<Output<MODE>>` @ https://github.com/stm32-rs/stm32f3xx-hal/pull/105/files#diff-73928a829df74085c37fe73a4f3484fcL184-L206

`PXx` is a fully erased pin type, different from `$PXx`.

src/rcc.rs
David-OConnor: Shouldn't be in this PR; meant it for a diff branch. Shouldn't be changed here.

src/gpio.rs
David-OConnor: Updated, I think.

David-OConnor: Updated, I think

ra-kete:
Please run rustfmt on the code again. It's currently complaining about one superfluous newline. This is what makes the Travis jobs fail.

ra-kete:
I also noticed that since the GPIO refactoring to use variants was merged, this should also be adapted in this PR. I've quickly done this myself, hope you don't mind me going off your changes @David-OConnor.

I had to create a new PR since I cannot just push to a branch in your repo: #114. Would be cool if you could check it out and report if it works for your use-case. You'll notice that I restricted the `InputPin` impls to `Output<OpenDrain>` only, based on what was discussed in #84. Looking at the onewire lib they are using OpenDrain too, so that shouldn't be an issue.

ra-kete:
Closed in favor of #114.

PR #107, opened 2020-06-17
Support 16 bit SPI

zklapow:
This adds support for sixteen bit words in SPI by adding a generic parameter for the wordsize (default to 8 for backwards compatibility). The `FullDuplex<u16>` impl is basically just copied with minor edits, there probably a better way to do this but copying seemed fine.

Ideally we could merge this alongside #99 and then SPI support would be much more complete!


ra-kete:
I'm not too happy about all that code duplication, we can do much better than that. I made you a PR with my suggested way of cleaning this up: zklapow/stm32f3xx-hal#1

zklapow:
I merged @ra-kete changes into this branch and brought this up to date with master, should be mergeable or at least ready for review now.

PR #117, opened 2020-07-20
improve rcc and gpio docs

TheZoq2:
This adds some documentation to the `rcc` and `gpio` documentation to make it clear how to aquire instances of their structs. Similar to what the f1 crate already has.

I don't have access to an f3 at the moment, so some of this may be wrong ;)


src/gpio.rs
```diff
@@ -1,4 +1,26 @@
...
+//!
+//! ```
+//! let dp = pac::Peripherals::take().unwrap();
+//! let rcc = dp.RCC.constrain();
+//!
+//! let mut gpioa = dp.GPIOA.split(&mut rcc.ahb);
+//! ```
+//!
+//! The resulting [Parts](gpioa/struct.Parts.html) struct contains one field for each
+//! pin, as well as some shared register.
```
strom-und-spiele, line 14: "as well as some shared register**s**, needed to configure the pins.

src/gpio.rs
```diff
@@ -1,4 +1,26 @@
...
+//!
+//! To use a pin, first use the relevant `into_...` function in the [pin](gpioa/struct.PA0.html).
+//!
+//! ```rust
+//! let pa0 = gpioa.pa0.into_push_pull_output(&mut gpioa.moder, &mut gpioa.otyper);
+//! ```
+//!
+//! And finally, you can use the functions from the [InputPin](../prelude/trait._embedded_hal_digital_InputPin.html) or [OutputPin](../prelude/trait._embedded_hal_digital_OutputPin.html) traits in
+//! `embedded_hal`
```
strom-und-spiele, line 23: I'd add a link to the toggle example
https://github.com/TheZoq2/stm32f3xx-hal/blob/doc_improvements/examples/toggle.rs

TheZoq2:
> I was wondering what one gets from reading those.

The way I tend to read the docs for HALs, and/or how I want people to read them, is to pick some feature they want to use and find a function that accomplishes that.

For example, someone wants to toggle a pin so they find the GPIO struct. The docs there tell you to call the `split` function, it takes `AHB` as input. However, the autogenerated docs don't tell you how to get your hand on `AHB`, which is what this adds.

In the F1 HAL, which I am much more familiar with, I kept running into the problem of not remembering where `MAPR` comes from. It sounds like it might be `RCC`, but it is not. For someone who isn't familiar with the crate at all, it would be even more difficult.

A "how to read the docs" section would be nice, but it would require people remembering where each struct comes from. Though perhaps one could link to that section instead of addding a bunch of `dp = pac::Peripherals.take` everywhere

src/gpio.rs
```diff
@@ -1,4 +1,26 @@
...
+//! let dp = pac::Peripherals::take().unwrap();
+//! let rcc = dp.RCC.constrain();
+//!
+//! let mut gpioa = dp.GPIOA.split(&mut rcc.ahb);
+//! ```
+//!
+//! The resulting [Parts](gpioa/struct.Parts.html) struct contains one field for each
+//! pin, as well as some shared register.
+//!
+//! To use a pin, first use the relevant `into_...` function in the [pin](gpioa/struct.PA0.html).
```
ra-kete, line 16:
```suggestion
//! To use a pin, first use the relevant `into_...` method of the [pin](gpioa/struct.PA0.html).
```

src/gpio.rs
```diff
@@ -1,4 +1,26 @@
...
+//!
+//! ```rust
+//! let pa0 = gpioa.pa0.into_push_pull_output(&mut gpioa.moder, &mut gpioa.otyper);
+//! ```
+//!
+//! And finally, you can use the functions from the [InputPin](../prelude/trait._embedded_hal_digital_InputPin.html) or [OutputPin](../prelude/trait._embedded_hal_digital_OutputPin.html) traits in
```
ra-kete, line 22: Could you make the lines adhere to some maximum length? I like to use 80 characters, rustfmt uses 100 by default. Both would be fine by me. I'd just like to avoid having to right-scroll when reading a comment, since that reduces understandability.

ra-kete:
I see how the RCC docs are debatable. It would be nice if there was a better way to figure out from looking at the docs how a type can be constructed. And there is the potential issue of the docs becoming bloated and repetitive if we document that everywhere.

However, keep in mind that the stm32f3xx-hal is probably the one code base that most embedded Rust beginners interact with very early in their learning curve, since that's what's used by the Discovery book. As such I think it doesn't hurt to err on the side of having "too much" documentation. I would also argue that in general we are pretty far away from having too much documentation ;) As such I welcome this change.

src/gpio.rs
Sh3Rm4n: As this is markdown, "url-aliases" could be used as well, something like this could improve the readability, as these are rather long url's to have the sentence. 

```suggestion
//! And finally, you can use the functions from the [InputPin][] or [OutputPin][]traits in

[InputPin]: ../prelude/trait._embedded_hal_digital_InputPin.html
[OutputPin]: ../prelude/trait._embedded_hal_digital_OutputPin.html
```

src/gpio.rs
Sh3Rm4n: Are examples even included in the `cargo doc` build. IIRC, this is not the case and it would result into a dead link, unfortunately.

Sh3Rm4n:
Doc-Improvments are always much appreciated. I feel the same way, that the repetitive comments in the rcc module feel more like fighting against the limitation of documentation generation. 

But these additions are rather small and I really appreciate these additions as it is really easy to get lost in the generated documentation. And RCC is the module **everyone** has to use, when working with this crate, which makes good and comprehensible documentation even more important.

src/gpio.rs
strom-und-spiele: You're right. 
https://github.com/rust-lang/cargo/issues/2760
I also don't think that using japaric's script is a good idea here.

TheZoq2:
Updated with the suggested length changes (though the last line is still quite long because of the URL.

For now, I added a link to the examples on this repo at the currently released version. This is what I do in the f1 hal as well, though it's not without flaws:

 - Requires updating on release. All the places where that's required can be found fairly easily using `grep v0.4.3` on every release though
- Will not be correct when people depend on the git version

src/rcc.rs
```diff
@@ -32,6 +32,13 @@ impl RccExt for RCC {
 }
 
 /// Constrained RCC peripheral
+///
+/// An instance of this struct is aquired by calling the [constrain](trait.RccExt.html#tymethod.constrain) function on the [pac::RCC](../pac/struct.RCC.html) struct.
```
ra-kete, line 36: Forgot one overlong line!

TheZoq2:
Oops, fixed :)

ra-kete:
Hey @TheZoq2, there are still two open suggestions, have you seen these? They are pretty nitpicky so I'd be fine with merging as is. Just wanted to make sure you didn't miss them.

TheZoq2:
I only see one open suggestion which I now fixed, did I miss the other?

PR #120, opened 2020-07-21
Use match instead of nested `if let`

Sh3Rm4n:
I stumbled upon this code again and thought I could make it a bit more readable.

### DIFF
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -438,32 +438,23 @@ impl CFGR {
     fn get_sysclk(&self) -> (u32, cfgr::SW_A, Option<PllConfig>) {
         // If a sysclk is given, check if the PLL has to be used,
         // else select the system clock source, which is either HSI or HSE.
-        if let Some(sysclk) = self.sysclk {
-            if let Some(hseclk) = self.hse {
-                if sysclk == hseclk {
-                    // No need to use the PLL
-                    // PLL is needed for USB, but we can make this assumption, to not use PLL here,
-                    // because the two valid USB clocks, 72 Mhz and 48 Mhz, can't be generated
-                    // directly from neither the internal rc (8 Mhz)  nor the external
-                    // Oscillator (max 32 Mhz), without using the PLL.
-                    (hseclk, cfgr::SW_A::HSE, None)
-                } else {
-                    let clock_with_pll = self.calc_pll(sysclk);
-                    (clock_with_pll.0, cfgr::SW_A::PLL, Some(clock_with_pll.1))
-                }
-            } else if sysclk == HSI {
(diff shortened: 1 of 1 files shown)


src/rcc.rs
```diff
@@ -438,32 +438,33 @@ impl CFGR {
     fn get_sysclk(&self) -> (u32, cfgr::SW_A, Option<PllConfig>) {
         // If a sysclk is given, check if the PLL has to be used,
         // else select the system clock source, which is either HSI or HSE.
-        if let Some(sysclk) = self.sysclk {
-            if let Some(hseclk) = self.hse {
-                if sysclk == hseclk {
+        match (self.sysclk, self.hse) {
+            (Some(sysclk), Some(hse)) => {
+                if sysclk == hse {
```
ra-kete, line 443: You could even put this in an if-guard on the match to remove one level of indentation

src/rcc.rs
```diff
@@ -438,32 +438,33 @@ impl CFGR {
...
+                    (sysclk, cfgr::SW_A::HSE, None)
                 } else {
-                    let clock_with_pll = self.calc_pll(sysclk);
-                    (clock_with_pll.0, cfgr::SW_A::PLL, Some(clock_with_pll.1))
+                    let (sysclk, pll_config) = self.calc_pll(sysclk);
+                    (sysclk, cfgr::SW_A::PLL, Some(pll_config))
+                }
+            }
+            (Some(sysclk), None) => {
+                if sysclk == HSI {
```
ra-kete, line 456: Same here, re if-guard

src/rcc.rs
```diff
@@ -438,32 +438,33 @@ impl CFGR {
...
+                    // No need to use the PLL
+                    (sysclk, cfgr::SW_A::HSE, None)
```
ra-kete, line 458: I haven't looked into the manual, but using `SW_A::HSE` here seems fishy. Shouldn't that be `SW_A::HSI`?

ra-kete:
Looks like we got another one of these hanging Travis builds here.

src/rcc.rs
Sh3Rm4n: Good idea!

src/rcc.rs
Sh3Rm4n: Hm. :thinking: It was there before (though it was my addition before as well). It should be `SW_A::HSI`. I'll recheck against the reference manual.

src/rcc.rs
Sh3Rm4n: Nice, with the help of if-guard, I could simplify it quite a bit further!

PR #133, opened 2020-09-01
Added DAC with basic features

David-OConnor:
- Includes builtin DAC trait that would [ideally be in embedded-hal](https://github.com/rust-embedded/embedded-hal/pull/247). Missing noise and tri wave features (commented-out and partially implemented) and triggers, but I think it's worth merging even without.

The 12-bit left mode doesn't work; probably needs bit arith on the value word integer.

edit: Added triggers and (probably broken) tri wave and noise.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,5 +1,5 @@
 [target.thumbv7em-none-eabihf]
-runner = "arm-none-eabi-gdb"
+runner = "arm-none-eabi-gdb -q -x openocd.gdb"
 rustflags = [
   "-C", "link-arg=-Tlink.x",
 ]
--- a/examples/dac.rs
+++ b/examples/dac.rs
@@ -3,9 +3,11 @@
 
 use panic_semihosting as _;
 
+use cortex_m::asm;
+
 use cortex_m_rt::entry;
 use stm32f3xx_hal::{
-    dac::{Dac, DacBits},
+    dac::{Dac, DacBits, Trigger},
     pac,
     prelude::*,
 };
@@ -28,13 +30,44 @@ fn main() -> ! {
     let dac_pin = gpioa.pa4.into_analog(&mut gpioa.moder, &mut gpioa.pupdr);
 
     // Set up the DAC
-    let mut dac = Dac::new(dp.DAC, dac_pin, DacBits::EightR, 3.3);
+    let reference_voltage = 2.915;
+    let mut dac = Dac::new(dp.DAC, dac_pin, DacBits::EightR, reference_voltage);
     dac.enable(&mut rcc.apb1);
 
-    dac.set_voltage(1.5);
-    // dac.set_value(128);  // or equivalently
+    // primitively wait about one second on a default stm32f3discovery
+    let wait = 8_000_000;
+    asm::delay(wait);
 
+    // directly set the output voltage to 1.5 V (valid values: 0..=reference_voltage)
+    dac.set_voltage(1.5);
+    asm::delay(3 * wait);
+
+    // equivalently:
+    dac.set_value(256 / 2);
+
+    // set the output off for 3 seconds
+    dac.set_value(0);
+    asm::delay(3 * wait);
+
+    // setup a trigger. From now on, setting a value or voltage will be stored but not converted
+    dac.set_trigger(Trigger::SoftwareTrigger);
+
+    // this is never converted:
+    dac.set_voltage(1.5);
+    asm::delay(wait);
+
+    // toggle the dac value every 2 seconds
     loop {
-        continue;
+        dac.set_value(0);
+        asm::delay(wait);
+
+        dac.trigger_software_trigger();
+        asm::delay(wait);
+
+        dac.set_value(255);
+        asm::delay(wait);
+
+        dac.trigger_software_trigger();
+        asm::delay(wait);
     }
 }
--- a/examples/dac.rs
+++ b/examples/dac.rs
@@ -4,42 +4,37 @@
 use panic_semihosting as _;
 
 use cortex_m_rt::entry;
-use embedded_hal::blocking::delay::DelayMs;
 use stm32f3xx_hal::{
-    dac::{Channel, Dac, DacBits},
-    delay::Delay,
+    dac::{Dac, DacBits},
     pac,
     prelude::*,
 };
 
 #[entry]
 fn main() -> ! {
-    // Set up CPU peripherals
-    let cp = cortex_m::Peripherals::take().unwrap();
-
     // Set up microcontroller peripherals
     let dp = pac::Peripherals::take().unwrap();
 
     // Set up clocks
     let mut flash = dp.FLASH.constrain();
     let mut rcc = dp.RCC.constrain();
-    let clocks = rcc.cfgr.freeze(&mut flash.acr);
+    let _clocks = rcc.cfgr.freeze(&mut flash.acr);
 
-    let mut delay = Delay::new(cp.SYST, clocks);
-
-    // Set up gpio pins if required
+    // Set up the DAC output pin.
+    // The output pin you choose selects the respective DAC channel:
+    //  - PA4: channel 1
+    //  - PA5: channel 2
     let mut gpioa = dp.GPIOA.split(&mut rcc.ahb);
-
-    // Set up the DAC
     let dac_pin = gpioa.pa4.into_analog(&mut gpioa.moder, &mut gpioa.pupdr);
 
-    let mut dac = Dac::new(dp.DAC, dac_pin, Channel::One, DacBits::EightR, 3.3);
+    // Set up the DAC
+    let mut dac = Dac::new(dp.DAC, dac_pin, DacBits::EightR, 3.3);
     dac.enable(&mut rcc.apb1);
 
     dac.set_voltage(1.5);
     // dac.set_value(128);  // or equivalently
 
     loop {
-        delay.delay_ms(1_000_u16);
+        continue;
     }
 }
--- a/examples/dac.rs
+++ b/examples/dac.rs
@@ -1,0 +1,45 @@
+#![no_main]
+#![no_std]
+
+use panic_semihosting as _;
+
+use cortex_m_rt::entry;
+use embedded_hal::blocking::delay::DelayMs;
+use stm32f3xx_hal::{
+    dac::{Channel, Dac, DacBits},
+    delay::Delay,
+    pac,
+    prelude::*,
+};
+
+#[entry]
+fn main() -> ! {
+    // Set up CPU peripherals
+    let cp = cortex_m::Peripherals::take().unwrap();
+
+    // Set up microcontroller peripherals
+    let dp = pac::Peripherals::take().unwrap();
+
+    // Set up clocks
+    let mut flash = dp.FLASH.constrain();
+    let mut rcc = dp.RCC.constrain();
+    let clocks = rcc.cfgr.freeze(&mut flash.acr);
+
+    let mut delay = Delay::new(cp.SYST, clocks);
+
+    // Set up gpio pins if required
+    let mut gpioa = dp.GPIOA.split(&mut rcc.ahb);
+
+    // Set up the DAC
+    let dac_pin = gpioa.pa4.into_analog(&mut gpioa.moder, &mut gpioa.pupdr);
+
+    let mut dac = Dac::new(dp.DAC, dac_pin, Channel::One, DacBits::EightR, 3.3);
+    dac.enable(&mut rcc.apb1);
+
+    dac.set_voltage(1.5);
+    // dac.set_value(128);  // or equivalently
+
+    loop {
+        delay.delay_ms(1_000_u16);
+    }
+}
--- a/src/dac.rs
+++ b/src/dac.rs
@@ -155,44 +155,6 @@ impl Dac {
             }
         });
     }
-
-    /// Independent trigger with single LFSR generation
-    /// See f303 Reference Manual section 16.5.2
-    pub fn trigger_lfsr(&mut self, trigger: Trigger, data: u32) {
-        // todo: This may not be correct.
-        self.regs.cr.modify(|_, w| match self.channel {
-            Channel::One => {
-                w.wave1().noise();
-                w.mamp1().bits(0b01)
-            }
-            Channel::Two => {
-                w.wave2().noise();
-                w.mamp2().bits(0b01)
-            }
-        });
-
-        self.set_trigger(trigger);
-        self.set_value(data);
-    }
-
-    /// Independent trigger with single triangle generation
-    /// See f303 Reference Manual section 16.5.2
-    pub fn trigger_triangle(&mut self, trigger: Trigger, data: u32) {
-        // todo: This may not be correct.
-        self.regs.cr.modify(|_, w| match self.channel {
-            Channel::One => {
-                w.wave1().triangle();
-                w.mamp1().bits(0b10)
-            }
-            Channel::Two => {
-                w.wave2().triangle();
-                w.mamp2().bits(0b10)
-            }
-        });
-
-        self.set_trigger(trigger);
-        self.set_value(data);
-    }
 }
 
 pub struct DacError {}
--- a/src/dac.rs
+++ b/src/dac.rs
@@ -46,37 +46,29 @@ pub enum DacBits {
 }
 
 #[derive(Clone, Copy)]
-/// Select a trigger, used by some features.
+/// Select an external event to trigger the DAC.
 pub enum Trigger {
-    /// Timer 6
-    Tim6,
-    /// Timers 3 or 8
-    Tim3_8,
-    /// Timer 7
-    Tim7,
-    /// Timer 15
-    Tim15,
-    /// Timer 2
-    Tim2,
-    /// Timer 4
-    Tim4,
-    /// Eg, for interrupts
+    Timer2,
+    Timer3Or8,
+    Timer4,
+    Timer6,
+    Timer7,
+    Timer15,
     Exti9,
-    /// A software trigger
-    Swtrig,
+    SoftwareTriggger,
 }
 
 impl Trigger {
     pub fn bits(&self) -> u8 {
         match self {
-            Self::Tim6 => 0b000,
-            Self::Tim3_8 => 0b001,
-            Self::Tim7 => 0b010,
-            Self::Tim15 => 0b011,
-            Self::Tim2 => 0b100,
-            Self::Tim4 => 0b101,
+            Self::Timer6 => 0b000,
+            Self::Timer3Or8 => 0b001,
+            Self::Timer7 => 0b010,
+            Self::Timer15 => 0b011,
+            Self::Timer2 => 0b100,
+            Self::Timer4 => 0b101,
             Self::Exti9 => 0b110,
-            Self::Swtrig => 0b111,
+            Self::SoftwareTriggger => 0b111,
         }
     }
 }
--- a/src/dac.rs
+++ b/src/dac.rs
@@ -59,7 +59,7 @@ pub enum Trigger {
 }
 
 impl Trigger {
-    pub fn bits(&self) -> u8 {
+    fn bits(&self) -> u8 {
         match self {
             Self::Timer6 => 0b000,
             Self::Timer3Or8 => 0b001,
@@ -100,15 +100,18 @@ impl Dac {
     /// Enable the DAC.
     pub fn enable(&mut self, apb1: &mut APB1) {
         apb1.enr().modify(|_, w| w.dac1en().enabled());
-        match self.channel {
-            Channel::One => self.regs.cr.modify(|_, w| w.en1().enabled()),
-            Channel::Two => self.regs.cr.modify(|_, w| w.en2().enabled()),
-        }
+        self.regs.cr.modify(|_, w| match self.channel {
+            Channel::One => w.en1().enabled(),
+            Channel::Two => w.en2().enabled(),
+        });
     }
 
     /// Disable the DAC
     pub fn disable(&mut self, apb1: &mut APB1) {
-        self.regs.cr.modify(|_, w| w.en1().disabled());
+        self.regs.cr.modify(|_, w| match self.channel {
+            Channel::One => w.en1().disabled(),
+            Channel::Two => w.en2().disabled(),
+        });
         apb1.enr().modify(|_, w| w.dac1en().disabled());
     }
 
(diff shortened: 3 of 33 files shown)


David-OConnor:
I'm not sure why it's failing the Travis. [Eg pac::DAC exists](https://docs.rs/stm32f3xx-hal/0.5.0/stm32f3xx_hal/pac/struct.DAC.html).

David-OConnor:
Added a modified version of https://github.com/stm32-rs/stm32f3xx-hal/pull/93, with many of the review requests added. (Will add more later).

That module is badly in need of getting added. I understand it's not perfect, but it's good enough. I got rid of LSI/HSE support as requested, to avoid exposing the extra RCC regs, per the discussion.

ra-kete:
Could you please split the DAC and RTC implementations into two self-contained PRs? This will make it easier to keep track of these two features and increase the chances that somebody finds time to review one or both of them.

David-OConnor:
The other PR showed no progress, and I accidentally committed the rtc file and modified rcc file since I use it in my projects. I would be thrilled if that were merged.

I added most of your review changes to that PR, but don't know how to justify the unsafe blocks. Would you be willing to merge that PR if it were modified to reflect what I have in this one? How would you like to proceed?

ra-kete:
I would suggest you open a new PR that duplicates #93 with the RTC-related changes from here. Make sure all CI checks pass, then I'll review it and if it looks good, we can merge and close #93.

And from this PR here, just remove all the RTC-related changes (by rewriting the history and force-pushing) so it contains only the DAC feature. And then the same procedure: Make sure the CI tests pass, the maintainers will review, and if everything looks good we'll merge it.

Hope that clears it up.

David-OConnor:
Done. [RTC pr](https://github.com/stm32-rs/stm32f3xx-hal/pull/136)

David-OConnor:
Any idea why it's failing CI? `##[error]Unable to create clippy annotations! Reason: HttpError: Resource not accessible by integration`.

ra-kete:
That error is related to the Github Action we are using (see https://github.com/actions-rs/clippy-check#limitations) and is not the reason clippy fails, so you can ignore it.

The clippy error is:
```
error: unused import: `panic_semihosting`
 --> examples/dac.rs:4:5
  |
4 | use panic_semihosting;
  |     ^^^^^^^^^^^^^^^^^
  |
  = note: `-D unused-imports` implied by `-D warnings`


error: this import is redundant
```

Replace that line with:
```
use panic_semihosting as _;
```

David-OConnor:
Can anyone figure out why this line doesn't work, but its equivalent for right-hand work? It's probably bit arithmetic I'm missing.
```rust
DacBits::TwelveL => self.regs.dhr12l1.modify(|_, w| unsafe { w.bits(val) }),
```

David-OConnor:
Moved away from trait-based approach, since I don't think it's general enough to make it into EH.

src/dac.rs
```diff
@@ -0,0 +1,220 @@
...
+            vref,
+        }
+    }
+
+    /// Enable the DAC.
+    pub fn enable(&mut self, apb1: &mut APB1) {
+        match self.id {
+            DacId::One => {
+                apb1.enr().modify(|_, w| w.dac1en().set_bit());
+                self.regs.cr.modify(|_, w| w.en1().set_bit());
```
ra-kete, line 96: The PAC usually provides more readable methods then `set_bit`/`clear_bit`. Using these makes the code more readable and can even avoid bugs. In this case, the readable method would be `.enabled()`.

src/dac.rs
```diff
@@ -0,0 +1,220 @@
...
+            DacId::One => {
+                self.regs.cr.modify(|_, w| w.ten1().set_bit());
+                self.regs
+                    .cr
+                    .modify(|_, w| unsafe { w.tsel1().bits(trigger.bits()) });
+            }
+            DacId::Two => {
+                self.regs.cr.modify(|_, w| w.ten2().set_bit());
+                self.regs.cr.modify(|_, w| w.tsel2().bits(trigger.bits()));
+            }
```
ra-kete, line 158: Similar to my comment above: I believe this sets a trigger for channel two of DAC1, not a trigger for DAC2

src/dac.rs
```diff
@@ -0,0 +1,220 @@
...
+        }
+    }
+
+    /// Independent trigger with single LFSR generation
+    /// See f303 Reference Manual section 16.5.2
+    pub fn trigger_lfsr(&mut self, trigger: Trigger, data: u32) {
+        // todo: This may not be correct.
```
ra-kete, line 165: I would rather not merge code that might not be correct. Do we have a way to verify this? If not we should maybe just remove it.

src/dac.rs
```diff
@@ -0,0 +1,220 @@
...
+            }
+        }
+        self.set_trigger(trigger);
+        self.set_value(data);
+    }
+
+    /// Independent trigger with single triangle generation
+    /// See f303 Reference Manual section 16.5.2
+    pub fn trigger_triangle(&mut self, trigger: Trigger, data: u32) {
+        // todo: This may not be correct.
```
ra-kete, line 183: See previous comment.

src/dac.rs
```diff
@@ -0,0 +1,220 @@
...
+                self.regs.cr.modify(|_, w| unsafe { w.wave2().bits(0b10) });
+                self.regs.cr.modify(|_, w| w.mamp2().bits(0b10));
+            }
+        }
+        self.set_trigger(trigger);
+        self.set_value(data);
+    }
+}
+
+impl fmt::Debug for Dac {
```
ra-kete, line 199: We usually don't impl the `Debug` trait for peripherals in the HAL. What's the reason for doing it in this case?

src/dac.rs
```diff
@@ -0,0 +1,220 @@
...
+            .field("bits", &self.bits)
+            .field("vret", &self.vref)
+            .finish()
+    }
+}
+
+pub struct DacError {}
+
+impl SingleChannelDac for Dac {
+    type Error = DacError;
```
ra-kete, line 212: For errors that are never instantiated we have `core::convert::Infallible`.

ra-kete (changes requested):

src/dac.rs
```diff
@@ -0,0 +1,248 @@
...
+    bits: DacBits,
+    vref: f32,
+}
+
+impl Dac {
+    /// Create a new DAC instance
+    pub fn new<P: Pins>(
+        regs: pac::DAC,
+        _pins: P,
+        channel: Channel,
```
ra-kete, line 109: Now it is possible for the user to pass in the wrong pin for the given channel. Instead can we select the channel based on the pin? Could be done via trait methods on `Pins`, to allow you to do `P::enable_dac_channel()` and `P::disable_dac_channel()` with the correct channel.

src/dac.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    /// Enable the DAC.
+    pub fn enable(&mut self, apb1: &mut APB1) {
+        match self.channel {
+            Channel::One => {
+                apb1.enr().modify(|_, w| w.dac1en().enabled());
+                self.regs.cr.modify(|_, w| w.en1().enabled());
+            }
+            Channel::Two => {
+                apb1.enr().modify(|_, w| w.dac2en().enabled());
```
ra-kete, line 129: I assume in your last changes you simply dropped support for DAC2? That's reasonable I think, considering the PAC doesn't have it. But you still enable DAC2 here. Should be `dac1en` in both cases, no?

src/dac.rs
ra-kete: Not sure why you marked this as resolved?

src/dac.rs
```diff
@@ -0,0 +1,248 @@
...
+    pub fn set_trigger(&mut self, trigger: Trigger) {
+        match self.channel {
+            Channel::One => {
+                self.regs.cr.modify(|_, w| w.ten1().set_bit());
+                self.regs
+                    .cr
+                    .modify(|_, w| unsafe { w.tsel1().bits(trigger.bits()) });
+            }
+            Channel::Two => {
+                self.regs.cr.modify(|_, w| w.ten2().set_bit());
```
ra-kete, line 186: The `tenX` fields also have `enabled()` setters, which are preferable to `set_bit()`

src/dac.rs
```diff
@@ -0,0 +1,248 @@
...
+    /// Output a constant signal, given a bit word.
+    fn try_set_value(&mut self, value: Self::Word) -> Result<(), Self::Error>;
+}
+
+/// This is an abstraction to ensure that the DAC output pin is configured
+/// as an analog output.
+pub trait Pins {}
+impl Pins for PA4<Analog> {}
+impl Pins for PA5<Analog> {}
+impl Pins for (PA4<Analog>, PA5<Analog>) {}
```
ra-kete, line 39: `PA4` is for channel 1 and `PA5` for channel 2, right? Your DAC constructor only allows selecting a single channel at a time. So why would one want to pass in both pins?

ra-kete (changes requested):

eldruin:
@ra-kete What is your intention/thoughts regarding this PR? We are thinking about merging the `SingleChannelDac` trait into `embedded-hal`. See: https://github.com/rust-embedded/embedded-hal/pull/247

ra-kete:
@eldruin I would like to merge the PR as soon as the remaining open review comments have been addressed.

eldruin:
@ra-kete Sounds good. I will hold the merge in `embedded-hal` then.

David-OConnor:
Ready for merge. This represents an improvement to the crate by adding missing functionality.

eldruin (commented):
The one TODO about correctness is still there.
I would also recommend squashing the changes before merging.

src/dac.rs
```diff
@@ -0,0 +1,233 @@
...
+    /// Disable the DAC
+    pub fn disable(&mut self, apb1: &mut APB1) {
+        match self.channel {
+            Channel::One => {
+                self.regs.cr.modify(|_, w| w.en1().disabled());
+                apb1.enr().modify(|_, w| w.dac1en().disabled());
+            }
+            Channel::Two => {
+                self.regs.cr.modify(|_, w| w.en2().disabled());
+                apb1.enr().modify(|_, w| w.dac2en().disabled());
```
ra-kete, line 128: This still disables DAC2 although we don't support it anymore. Like with `enable`, this should also changed to `dac1en`

src/dac.rs
```diff
@@ -0,0 +1,233 @@
...
+    Tim6,
+    /// Timers 3 or 8
+    Tim3_8,
+    /// Timer 7
+    Tim7,
+    /// Timer 15
+    Tim15,
+    /// Timger 2
+    Tim2,
+    /// Tim4 is Not available on DAC 2.
```
ra-kete, line 63: Since we don't support DAC2 anymore, the "not available" not here is not necessary either

ra-kete (changes requested):

ra-kete:
@eldruin No worries, I will squash anyway when I merge.

@David-OConnor Please don't just mark review comments as "resolved" without at least commenting why you think they can be ignored.

ra-kete:
> I did a squash, and now the `Files changed` tab is reflecting changes from other PRs.

A squash is simply the act of collecting multiple commits into a single one, it shouldn't pull in other changes. If you can't make it work, please just reset back to the state before. I'll do the squash when merging.
 
> I marked the comments resolved when I asked you questions and you didn't respond to them.

Okay, there seems to be some communication issue here. I am not aware of any questions you asked me. I normally would expect such questions as comments under my review comments. Where did you ask me?

David-OConnor:
Thanks - squash fixed! For the questions, they're in the [review comments](https://github.com/stm32-rs/stm32f3xx-hal/pull/133/files).

src/dac.rs
strom-und-spiele: If not really needed, try to avoid `fmt` alltogether, because [Formatting is Unreasonably Expensive for Embedded Rust](https://jamesmunns.com/blog/fmt-unreasonably-expensive/)

src/dac.rs
```diff
@@ -0,0 +1,220 @@
...
+pub enum Trigger {
+    Tim6,
+    Tim3_8,
+    Tim7,
+    Tim15,
+    Tim2,
+    Tim4, // Not available on DAC 2.
+    Exti9,
+    Swtrig,
+}
```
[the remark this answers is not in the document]
strom-und-spiele: Instead of
```Rust
pub enum Trigger {
  /// Timer 15
  Tim15
  /// A software trigger
  Swtrig
}
```
I'd love to see simply
```
pub Trigger {
  Timer15,
  SoftwareTrigger,
}
```
or something alike.
I also see the need to at least document that Timer 4 is not available on DAC 2. Even better would be to not allow it during compilation.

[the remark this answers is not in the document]
ra-kete: Since this PR doesn't include support for DAC2 anymore, I don't think it is necessary to specifically mention that Timer 4 is not available there. That would just confuse people.

I originally requested the per-variant annotations to document that Timer 4 edge-case. Since that now isn't relevant anymore, I agree that the comments don't add much value anymore either.

David-OConnor:
I recommend merging, then fixing the minutae in a separate PR.

ra-kete:
I think I already made clear my view that code we merge should be of high quality and production-ready. This crate has enough code that needs refactoring already, no reason to willingly add more.

I don't see why we can't just fix the few known issues in this PR? None of them look very difficult to me. If you don't feel up to it, let me know. I can fix them myself in this PR.

David-OConnor:
I quite frankly don't. I'm willing to make things work, but I'm not willing to play  this game any more.

ra-kete (approved):

ra-kete:
I've updated the PR to the best of my knowledge. @strom-und-spiele, @David-OConnor, would be good if you could look at it again to ensure I didn't mess up any of the finer DAC details.

strom-und-spiele:
LGTM. 
Making Commits based on topics made it easy to understand what you did each step. 
Selecting channels based on the Pin pulls down complexity and allows for ergonomic handling.
On Monday I'll set an oscilloscope up in my office to check triangle and LSFR generation. If I can make it work with the given code in a reasonable amount of time I'd vote for adding it to the example and keeping it.

Apart form that: thanks @David-OConnor and @ra-kete  for you work.

David-OConnor:
Looks good!

ra-kete:
@strom-und-spiele Okay, sounds good! I'm all for keeping this functionality if we can verify it. I'll hold off merging until you have the results then.

Sh3Rm4n (changes requested):
I think this is missing a `free()` method to give back ownership of the `Pin`, see [C-FREE](https://docs.rust-embedded.org/book/design-patterns/hal/interoperability.html#c-free)

strom-und-spiele:
Updated the required features for all examples that were still referencing `stm32f303` as this is conflicting with setting up the pins correctly.

Playtime for today is over. Couldn't make the current example work. This is the error I was stuck with last:
```
$ cargo run --example dac --features stm32f303xc
    Finished dev [unoptimized + debuginfo] target(s) in 0.11s
     Running `arm-none-eabi-gdb -q -x openocd.gdb target/thumbv7em-none-eabihf/debug/examples/dac`
Reading symbols from target/thumbv7em-none-eabihf/debug/examples/dac...
0x080065de in ?? ()
openocd.gdb:10: Error in sourced command file:
Dwarf Error: Cannot find DIE at 0x940 referenced from DIE at 0x8e5 [in module /home/bastian/tmp/stm32_dac/stm32f3xx-hal/target/thumbv7em-none-eabihf/debug/examples/dac]
```

ra-kete:
> I think this is missing a `free()` method to give back ownership of the `Pin`, see [C-FREE](https://docs.rust-embedded.org/book/design-patterns/hal/interoperability.html#c-free)

Good catch, I will add that. I think for this I'll have to make the `Dac` abstraction generic over the output pin.

@strom-und-spiele Looks like you're building in debug mode. Have you tried `--release`?

strom-und-spiele:
`--release` made the example build with an output of ~1.3 V. This might be b.c. the multimeter on my desk is not that precise, the input is too low or the reference voltage on the board is not 3V but lower. If I'm quick with the stuff I have to do for what I'm getting paid for here I'll try it again later today or tomorrow ;)

strom-und-spiele:
So I hooked it up to the osci and I get the following with a freshly out-of-the-box STM32F3DISCOVERY and v_ref =3.3 (as in the example code)
```
set_value   - measured Volts
128         - 1.49
255         - 2.915
256         - 0 (do we want this to happen?)

set_voltage - measured volts
1.0         - 0.9
1.5         - 1.3
3.0         - 2.7
3.3         - 2.915
5.0         - 1.5 (do we want this to happen?)
```
changed v_ref to 2.915 and got:
```
set_voltage - measured volts
1.0         - 1.0
1.5         - 1.5
2.5         - 2.5
5.0         - 2.1 (do we want this to happen?)
```
Need to rush home now but will try the dropped patterns tomorrow.

src/dac.rs
```diff
@@ -0,0 +1,171 @@
...
+impl Pin for PA4<Analog> {
+    const CHANNEL: Channel = Channel::One;
+}
+
+impl Pin for PA5<Analog> {
+    const CHANNEL: Channel = Channel::Two;
+}
+
+#[derive(Clone, Copy)]
+/// Select the channel
```
Sh3Rm4n, line 38: Really minor nit. In all other modules documentation string come before derive annotations.

src/dac.rs
```diff
@@ -0,0 +1,171 @@
...
+                DacBits::TwelveL => self.regs.dhr12l2.modify(|_, w| unsafe { w.bits(val) }),
+                DacBits::TwelveR => self.regs.dhr12r2.modify(|_, w| unsafe { w.bits(val) }),
+            },
+        }
+    }
+
+    /// Set the DAC voltage. `v` is in Volts.
+    pub fn set_voltage(&mut self, volts: f32) {
+        let val = match self.bits {
+            DacBits::EightR => ((volts / self.vref) * 255.) as u32,
```
Sh3Rm4n, line 140: Maybe there is a better, robust and explicit version (e.g. `u32::try_from(...)`). 
I'm not sure it is needed, though.

Sh3Rm4n (changes requested):

Sh3Rm4n:
> So I hooked it up to the osci and I get the following with a freshly out-of-the-box STM32F3DISCOVERY and v_ref =3.3 (as in the example code)
> 
> ```
> set_value   - measured Volts
> 128         - 1.49
> 255         - 2.915
> 256         - 0 (do we want this to happen?)
> ```

Seems like an u8 (256) bit overflow?

strom-und-spiele:
> > So I hooked it up to the osci and I get the following with a freshly out-of-the-box STM32F3DISCOVERY and v_ref =3.3 (as in the example code)
> > ```
> > set_value   - measured Volts
> > 128         - 1.49
> > 255         - 2.915
> > 256         - 0 (do we want this to happen?)
> > ```
> 
> Seems like an u8 (256) bit overflow?

sorry, I was in a rush and just c&p'ed the same phrase over and over. This is definitely an overflow issue. the implied question was whether the expected behavior is for the value to overflow. 
Say the ref voltage is 2V. If I set the output voltage to 3.3V the actual output voltage will be 1.3V. This might be what I wanted but it might as well be totally wrong. IMHO the best reaction would be to abort during compile time and force the user to apply a "mod 2" operation explicitly in the calling code.

Sh3Rm4n:
>  IMHO the best reaction would be to abort during compile time and force the user to apply a "mod 2" operation explicitly in the calling code.

As `set_voltage()` and `set_value()` can not fail they should not fail. Tricks like these should be included in the infallible version. 

If this overflow error is not preventable, e.g. value > v_ref, then `set_value()` and `set_voltage()` should fail.

src/dac.rs
```diff
@@ -0,0 +1,171 @@
...
+            },
+            Channel::Two => match self.bits {
+                DacBits::EightR => self.regs.dhr8r2.modify(|_, w| unsafe { w.bits(val) }),
+                DacBits::TwelveL => self.regs.dhr12l2.modify(|_, w| unsafe { w.bits(val) }),
+                DacBits::TwelveR => self.regs.dhr12r2.modify(|_, w| unsafe { w.bits(val) }),
+            },
+        }
+    }
+
+    /// Set the DAC voltage. `v` is in Volts.
```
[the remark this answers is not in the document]
strom-und-spiele: removed it.

strom-und-spiele:
Couldntt make the wave generation work on short notice. I advise merging without them. 

However I noticed that there was no way to actually trigger the software trigger, so I added it.

src/dac.rs
```diff
@@ -0,0 +1,179 @@
...
+            DacBits::EightR => ((volts / self.vref) * 255.) as u32,
+            DacBits::TwelveL => ((volts / self.vref) * 4_095.) as u32,
+            DacBits::TwelveR => ((volts / self.vref) * 4_095.) as u32,
+        };
+
+        self.set_value(val);
+    }
+
+    // Select and activate a trigger. See f303 Reference manual, section 16.5.4.
+    pub fn set_trigger(&mut self, trigger: Trigger) {
```
Sh3Rm4n, line 149: I wonder if we need a "`get_set_trigger`" function to check which trigger is set?
Not really needed, more like a nice to have.

src/dac.rs
```diff
@@ -0,0 +1,179 @@
...
+                unsafe { w.tsel1().bits(trigger.bits()) }
+            }
+            Channel::Two => {
+                w.ten2().enabled();
+                w.tsel2().bits(trigger.bits())
+            }
+        });
+    }
+
+    /// Takes the value stored via set_value or set_voltage and converts it to output on the Pin.
```
Sh3Rm4n, line 162: I did not understand this comment on the first read without context. 

And maybe add
```suggestion
    /// Takes the value stored via set_value or set_voltage and converts it to output on the Pin.
    /// 
    /// Does nothing, if software trigger is not enabled, see `set_trigger()`. 
```

.cargo/config
```diff
@@ -1,5 +1,5 @@
 [target.thumbv7em-none-eabihf]
-runner = "arm-none-eabi-gdb"
+runner = "arm-none-eabi-gdb -q -x openocd.gdb"
```
ra-kete, line 2: @strom-und-spiele Do you think it is a good idea to add a reference to a file that is not part of this repo?
IMO we probably shouldn't even specify GDB as a runner, as there are other options. But not sure what the intentions were when this was added.

examples/dac.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    // Set up the DAC output pin.
+    // The output pin you choose selects the respective DAC channel:
+    //  - PA4: channel 1
+    //  - PA5: channel 2
+    let mut gpioa = dp.GPIOA.split(&mut rcc.ahb);
+    let dac_pin = gpioa.pa4.into_analog(&mut gpioa.moder, &mut gpioa.pupdr);
+
+    // Set up the DAC
+    let reference_voltage = 2.915;
```
ra-kete, line 33: @strom-und-spiele Where is that voltage value coming from?

ra-kete: Ah, I see you got that from experimentation. Is this a value we can expect to be the same on each user of an F3DISCOVERY? If not, I think we should document how a user can find there own correct value. There must be some way to measure/infer the VREF for a given board, no?

src/dac.rs
ra-kete: Unless we have a convincing use-case for it now, I would leave it out for simplicity. We can always add it later if it's actually needed.

ra-kete:
I implemented/responded to all remaining review comments, except for the issue of handling invalid voltages/values.

There is not much we can do at compile time. We could limit setting an 8-bit value to a `u8`, but for setting 12-bit values Rust doesn't provide us a fitting data type. IMO it's better to handle both cases uniformly be returning a `Result`. Same is true for setting a voltage, I don't see what we could do at compile time to ensure no invalid value can be set.

So I propose making both `set_value` and `set_voltage` fallible. This would also work well with the `SingleChannelDac` trait, which already assumes that conversions can fail.

.cargo/config
strom-und-spiele: this was not intended.

examples/dac.rs
strom-und-spiele: I wasn't so sure either and checked the schematics of the stm32f3discovery again.

VREF is a stabilized version VDD.
VDD is directly connected via Jumper 3 to 3V
3V is powered by the 5V supply and adjusted by a [BAT60JFILM](https://www.mouser.de/datasheet/2/389/bat60-1848979.pdf) which has a voltage drop that depends on the power being used, so it's roughly 3V. Less if your chip is running something more demanding.

I suggest
* replacing the 2.9… by a solid 3
* add to the docs the info that the ref value should be known by the schematics and can be roughly determined by setting `set_value` to 255 and measuring the output
* add a hint to the docs in the example

strom-und-spiele:
> There is not much we can do at compile time. We could limit setting an 8-bit value to a `u8`, but for setting 12-bit values Rust doesn't provide us a fitting data type. IMO it's better to handle both cases uniformly be returning a `Result`. Same is true for setting a voltage, I don't see what we could do at compile time to ensure no invalid value can be set.

While nothing stops us from using [a twelfe bit type](https://github.com/martinmroz/12bit) I agree that uniformity is nicer here.

PR #136, opened 2020-09-03
Rtc v2

David-OConnor:
Based on https://github.com/stm32-rs/stm32f3xx-hal/pull/93, by @smedellin90

Has most of the review changes implemented, but is missing the unsafe justifications; I'm not sure how to do them.


src/rtc.rs
```diff
@@ -0,0 +1,448 @@
...
+
+        Ok(())
+    }
+
+    fn set_seconds(&mut self, seconds: u8) -> Result<(), Self::Error> {
+        if seconds > 59 {
+            return Err(Error::InvalidInputData);
+        }
+        let (st, su) = bcd2_encode(seconds as u32);
+        self.modify(|regs| regs.tr.write(|w| unsafe { w.st().bits(st).su().bits(su) }));
```
ra-kete, line 125: Using a `write` will reset all the fields not written. I believe in this case this means all other time fields (minutes, hours, ...) will be set to 0. I guess that is not the intended behavior?

Same for some of the other `set_*` methods below. I suggest you check all register `write`s again and make sure they shouldn't actually be `modify`s.

src/rtc.rs
```diff
@@ -0,0 +1,448 @@
...
+        let (mnt, mnu) = bcd2_encode(minutes as u32);
+        self.modify(|regs| {
+            regs.tr
+                .write(|w| unsafe { w.mnt().bits(mnt).mnu().bits(mnu) })
+        });
+
+        Ok(())
+    }
+
+    fn set_hours(&mut self, hours: rtcc::Hours) -> Result<(), Self::Error> {
```
ra-kete, line 143: Throughout this module you often use `rtcc::Hours`, even though `Hours` is already imported into the module scope at the beginning, so the `rtcc::` prefix is redundant. I suggest you remove it everywhere to be consistent.

src/rtc.rs
```diff
@@ -0,0 +1,448 @@
...
+        let (ht, hu) = hours_to_register(hours)?;
+        match hours {
+            Hours::H24(_h) => {
+                self.set_24h_fmt();
+                self.modify(|regs| regs.tr.write(|w| unsafe { w.ht().bits(ht).hu().bits(hu) }));
+            }
+            Hours::AM(_h) | Hours::PM(_h) => {
+                self.set_12h_fmt();
+                self.modify(|regs| regs.tr.write(|w| unsafe { w.ht().bits(ht).hu().bits(hu) }));
```
ra-kete, line 152: This line is the same in both cases, so you can move it under the match instead

src/rtc.rs
```diff
@@ -0,0 +1,448 @@
...
+        let (yt, yu) = bcd2_encode(year as u32);
+        self.modify(|regs| regs.dr.write(|w| unsafe { w.yt().bits(yt).yu().bits(yu) }));
+
+        Ok(())
+    }
+
+    /// set date using NaiveDate (ISO 8601 calendar date without timezone)
+    /// WeekDay is set using set_weekday method
+    fn set_date(&mut self, date: &NaiveDate) -> Result<(), Self::Error> {
+        let (yt, yu) = bcd2_encode((date.year() - 1970) as u32);
```
ra-kete, line 204: If `date.year()` is less than 1970, this will underflow. You should check that and return an `InvalidInputData` error in this case

src/rtc.rs
```diff
@@ -0,0 +1,448 @@
...
+                    .yu()
+                    .bits(yu)
+            });
+        });
+
+        Ok(())
+    }
+
+    fn set_datetime(&mut self, date: &NaiveDateTime) -> Result<(), Self::Error> {
+        // Check if unsigned integer affects encoding to bcd
```
ra-kete, line 229: What does this comment mean?

src/rtc.rs
```diff
@@ -0,0 +1,448 @@
...
+
+        self.set_24h_fmt();
+        let (yt, yu) = bcd2_encode((date.year() - 1970) as u32);
```
ra-kete, line 232: This can also underflow

src/rtc.rs
```diff
@@ -0,0 +1,448 @@
...
+        });
+
+        Ok(())
+    }
+
+    fn get_seconds(&mut self) -> Result<u8, Self::Error> {
+        let tr = self.regs.tr.read();
+        let seconds = bcd2_decode(tr.st().bits(), tr.su().bits());
+        Ok(seconds as u8)
+    }
```
ra-kete, line 282: Two general observations for all these `get_*` methods here:
- They never return an error, so there isn't a reason for a `Result` return type, right?
- Why do they take `&mut self` instead of `&self`? These are getters, so I wouldn't expect them to mutate anything

src/rtc.rs
```diff
@@ -0,0 +1,448 @@
...
+// Two 32-bit registers (RTC_TR and RTC_DR) contain the seconds, minutes, hours (12- or 24-hour format), day (day
+// of week), date (day of month), month, and year, expressed in binary coded decimal format
+// (BCD). The sub-seconds value is also available in binary format.
+//
+// The following helper functions encode into BCD format from integer and
+// decode to an integer from a BCD value respectively.
+fn bcd2_encode(word: u32) -> (u8, u8) {
+    (
+        (word / 10).try_into().unwrap(),
+        (word % 10).try_into().unwrap(),
```
ra-kete, line 378: I think instead of panicking we should return a `Result<(u8, u8), Error>` here with an `InvalidInputData` if the input is too large.

I think in the old PR I had advocated to making this check panicking, but I've since been convinced that it is better to limit panics in HAL code, as much as possible.

src/rtc.rs
```diff
@@ -0,0 +1,448 @@
...
+        Hours::AM(h) => Ok(bcd2_encode((h - 1) as u32)),
+        Hours::PM(h) => Ok(bcd2_encode((h + 11) as u32)),
+    }
+}
+
+fn hours_to_u8(hours: rtcc::Hours) -> u8 {
+    if let rtcc::Hours::H24(h) = hours {
+        h
+    } else {
+        panic!("hours could not be destructured into rtc::Hours::H24(h)");
```
ra-kete, line 398: Here too, think about returning an error instead

src/rtc.rs
```diff
@@ -0,0 +1,448 @@
...
+    }
+}
+
+/// Enable the low frequency external oscillator. This is the only mode currently
+/// supported, to avoid
```
ra-kete, line 403: This comment seems incomplete.

src/rtc.rs
```diff
@@ -0,0 +1,448 @@
...
+fn enable_lse(bdcr: &mut BDCR, bypass: bool) {
```
ra-kete, line 404: Since this function is only called once with `bypass` set to `false`, I wouldn't make this an argument and just hardcode it in the register write

src/rtc.rs
```diff
@@ -0,0 +1,448 @@
...
+    bdcr.bdcr()
+        .modify(|_, w| w.lseon().set_bit().lsebyp().bit(bypass));
+    while bdcr.bdcr().read().lserdy().bit_is_clear() {}
+}
+
+fn unlock(apb1: &mut APB1) {
+    let pwr = unsafe { &(*PWR::ptr()) };
```
ra-kete, line 411: This doesn't look safe to me. There is nothing to ensure we have exclusive access to the PWR register here.
Easiest way to solve this would be to pass in PWR as a `&mut`, same as APB1.

src/rtc.rs
```diff
@@ -0,0 +1,448 @@
...
+        if (weekday < 1) || (weekday > 7) {
+            return Err(Error::InvalidInputData);
+        }
+        self.modify(|regs| regs.dr.write(|w| unsafe { w.wdu().bits(weekday) }));
+
+        Ok(())
+    }
+
+    fn set_day(&mut self, day: u8) -> Result<(), Self::Error> {
+        if (day < 1) | (day > 31) {
```
ra-kete, line 169: Pretty sure you want to use logical OR here (`||`), not bitwise OR

src/rtc.rs
```diff
@@ -0,0 +1,448 @@
...
+            // RTC clock source selection
+            .rtcsel()
+            .bits(LSE_BITS)
+            // Enable RTC
+            .rtcen()
+            .set_bit()
+            // RTC backup Domain reset bit set low
+            .bdrst()
+            .clear_bit()
+    });
```
ra-kete, line 447: The BDCR fields are documented in the PAC, which means they have convenience functions that can make this more readable:

```rust
    bdcr.bdcr().modify(|_, w| w.bdrst().enabled());
    bdcr.bdcr().modify(|_, w| {
        w.rtcsel().lse();
        w.rtcen().enabled();
        w.bdrst().disabled()
    });
```

It's a good idea to use those special accessor functions whenever they are available as it helps readability

ra-kete (changes requested):
Thanks for your effort! Most of my comments should be pretty easy for you to fix, even though I realize many apply to code not originally written by you.

src/rtc.rs
David-OConnor: Added `bypass` to `new` constructor. It's required if you're using a standalone oscillator.

src/rtc.rs
```diff
@@ -0,0 +1,448 @@
...
+                    .mnt()
+                    .bits(mnt)
+                    .mnu()
+                    .bits(mnu)
+                    .st()
+                    .bits(st)
+                    .su()
+                    .bits(su)
+                    .pm()
+                    .clear_bit()
```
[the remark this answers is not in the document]
David-OConnor: I thought that format was weird too; didn't understand it, or why the closures were nested.

src/rtc.rs
David-OConnor: This is due to the more general specifications of the RTCC trait

ra-kete: Ah yes, I missed that you are implementing a trait here.

David-OConnor:
All reviews addressed.

ra-kete (changes requested):
Looks good! Just a small suggestion, I think some of these `modify`s should actually be `write`s now. I marked them with comments.

src/rtc.rs
```diff
@@ -0,0 +1,428 @@
...
+    type Error = Error;
+
+    /// set time using NaiveTime (ISO 8601 time without timezone)
+    /// Hour format is 24h
+    fn set_time(&mut self, time: &NaiveTime) -> Result<(), Self::Error> {
+        self.set_24h_fmt();
+        let (ht, hu) = bcd2_encode(time.hour())?;
+        let (mnt, mnu) = bcd2_encode(time.minute())?;
+        let (st, su) = bcd2_encode(time.second())?;
+        self.regs.tr.modify(|_, w| unsafe {
```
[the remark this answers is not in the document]
David-OConnor: Done

smedellin90:
@David-OConnor , thank you for getting around to cleaning up and improving the code!
@ra-kete , thank you for the reviewing and your patience!

ra-kete (approved):

ra-kete:
LGTM. Can you add a changelog entry too? Sorry, forgot about that before.

David-OConnor:
Done. @smedellin90 : Thanks for making this. I've been relying on it for months now!

PR #137, opened 2020-09-04
Migrate to traits from embedded-dma crate

ra-kete:
This PR removes the DMA traits from the HAL code and instead replaces them with the ones from the [embedded-dma crate](https://crates.io/crates/embedded-dma). Those traits are basically the same, so this is only a code move, no functional changes.


dfrankland (approved):
I dunno a lot about DMA, but the code looks good 😁

PR #148, opened 2020-09-27
Added low power module

David-OConnor:
Allows entering Sleep, Stop, or Standby modes by calling a function.


David-OConnor:
Thoughts on why CI is failing for rustfmt?

ra-kete:
> Thoughts on why CI is failing for rustfmt?

If you click on the details of the failed job, you can see rustfmt's suggested diff. In this case, some imports have the wrong order. You can also just run rustfmt on you code locally before committing to prevent something like this in the future (same for clippy).

David-OConnor:
Thanks. I think my local fmt is set up differently.

David-OConnor:
Please merge

PR #152, opened 2020-10-14
Remove divisor check for HSE PRE_DIV

### LINKED ISSUE #151: Cannot calculate PLL for stm32f303xc with 8MHz HSE running at maximum speed (72MHz)
Hi, I was having trouble setting a 72MHz sysclk with an 8MHz HSE. 
```
let _clocks = rcc
    .cfgr
    .use_hse(8.mhz())
    .sysclk(72.mhz())
    .hclk(72.mhz())
    .pclk1(36.mhz())
    .pclk2(72.mhz())
    .freeze(&mut flash.acr);
```
Through a little bit of debugging I found out `fn calc_pll` might not be deriving the correct divisor and multiplier for this configuration, causing an assertion to fail.
```
#[cfg(not(any(
        feature = "stm32f302xd",
        feature = "stm32f302xe",
        feature = "stm32f303xd",
        feature = "stm32f303xe",
        feature = "stm32f398"
    )))]
    fn calc_pll(&self, sysclk: u32) -> (u32, PllConfig) {
        let pllsrcclk = self.hse.unwrap_or(HSI / 2);
        // Get the optimal value for the pll divisor (PLL_DIV) and multiplier (PLL_MUL)
        // Only for HSE PLL_DIV can be changed
        let (pll_mul, pll_div): (u32, Option<u32>) = if self.hse.is_some() {
            // Get the optimal value for the pll divisor (PLL_DIV) and multiplier (PLL_MUL)
            // with the greatest common divisor calculation.
            let common_divisor = gcd(sysclk, pllsrcclk);
            let mut multiplier = sysclk / common_divisor;  // <<<<<<<< multiplier = 9
            let mut divisor = pllsrcclk / common_divisor;  // <<<<<<<< divisor = 1

            // Check if the multiplier can be represented by PLL_MUL
            // or if the divisor can be represented by PRE_DIV
            if multiplier == 1 || divisor == 1 {           // <<<<<<<< multiplier & divisor gets doubled
                // PLL_MUL minimal value is 2
                multiplier *= 2;
                // PRE_DIV minimal value is 2
                divisor *= 2;
            }

            // PLL_MUL maximal value is 16
            assert!(divisor <= 16);
            // PRE_DIV maximal value is 16
            assert!(multiplier <= 16);                     // <<<<<<<< since multiplier = 18, this assertion fails

            (multiplier, Some(divisor))
        }
```
For reference, here's a clock configuration generated by STM32CudeIDE:
Capture

Here are some of my guesses, but since I'm new to this project and embedded in general, please take what I say with a grain of salt:
  - Doubling the divisor and multiplier feel wrong to me.
    - For my particular case (8MHz HSE + 72MHz sysclk), the ratio `9/1` is perfectly representable yet `calc_pll` rejects it by making it `18/2`. The comments say the minimal value for both `PLL_MUL` and `PRE_DIV` is 2, but STM32CudeIDE uses `RCC_HSE_PREDIV_DIV1` and `RCC_PLL_MUL9` in the code it generates and since `PREDIV_A` does have a `DIV1` variant, should we try doubling only when the multiplier is 1?
    - For other configurations that currently work, it's still a little bit weird. For example, for an 8MHz HSE + 48MHz sysclk, `calc_pll` represents the ratio as `12/2` instead of `6/1`.
  - I wonder if what we really need to check here is "whether the ratio can be represented as `N/M` where `N` and `M` are integers such that `2 <= N <= 16` and `1 <= M <= 16`". Although it's unlikely this alone is sufficient for a valid clock configuration.

Anyway, I'd be happy to put up a PR if someone could confirm my guesses or point me in the right direction.

Sh3Rm4n:
Fixes #151

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Introduced auto-generated GPIO mappings based on the STM32CubeMX database
   ([#129](https://github.com/stm32-rs/stm32f3xx-hal/pull/129))
 
+- Fixed [#151][] not being 
+  able to generate 72 MHz HCLK for stm32f303xc devices ([#152](https://github.com/stm32-rs/stm32f3xx-hal/pull/152))
+
+[#151]: https://github.com/stm32-rs/stm32f3xx-hal/issues/151
+
 ## [v0.5.0] - 2020-07-21
 
 ### Added
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -369,11 +369,9 @@ impl CFGR {
             let mut divisor = pllsrcclk / common_divisor;
 
             // Check if the multiplier can be represented by PLL_MUL
-            // or if the divisor can be represented by PRE_DIV
-            if multiplier == 1 || divisor == 1 {
+            if multiplier == 1 {
                 // PLL_MUL minimal value is 2
                 multiplier *= 2;
-                // PRE_DIV minimal value is 2
                 divisor *= 2;
             }
 


ra-kete (approved):
LGTM

PR #154, opened 2020-10-19
Support I2C transfers of more than 255 bytes

Piroro-hs:
This adds support for I2C write and read of more than 255 bytes, and I2C write of 0 byte.


Sh3Rm4n (approved):
LGTM! Thanks for the contribution. 

Can you please add an entry to the changelog? :)

Piroro-hs:
Changelog updated!

Sh3Rm4n:
Thanks! :)

PR #156, opened 2020-10-22
Add support for HSE bypass and CSS

### LINKED ISSUE #142: Add clock HSE bypass
If you use a MEMs oscillator etc, you can bypass the oscillator outputs and use them as GPIOs. 

Eg does this:
```rust
rcc.cr.modify(|_, w| w.hsebyp().bit(true));
```

With this API:
```rust
rcc.set_hse_bypass(true);
```

Present workaround to deal with private or missing registers in `rcc.rs`:
```rust
unsafe { (*pac::RCC::ptr()).cr.modify(|_, w| w.hsebyp().bit(true)) }
```

Or maybe you add it to the `rcc.cfgr` chaining pattern:
```rust
    let clocks = rcc
        .cfgr
        .use_hse(8.mhz()) // Use the high-speed external oscillator
        // .hsclk(8.mhz())  // Sets a frequency for the AHB bus
        .pclk1(24.mhz()) // Sets a frequency for the APB1 bus
        // .pclk2(24.mhz ())  // Sets a frequency for the APB2 bus
        .sysclk(48.mhz()) // Sets the system (core) frequency
        // 16-72 Mhz valid range for for PLL. How do we set that?
        .hse_bypass()
        .freeze(&mut flash.acr);
```

Currently have to use raw pointers due to bdcr being crate-level private, and cr not being exposed at all. I'm on too many forks right now to make this a PR until the timer and DAC ones are closed. Note that in [this RTC PR](https://github.com/stm32-rs/stm32f3xx-hal/pull/93), we skipped some features to avoid CR being exposed. This use case is a second vote for exposing CR on a pub(crate) level as we did bdcr in there.

LSE bypass works in a similar way, and is already handled in the `rtc` module.

Piroro-hs:
Fix #142.

However, APIs for the same thing are vary in all stm32-rs HALs... Are there any suggestions?

* this PR
```rust
let clocks = rcc.cfgr/* .use_hse_crystal(8.mhz()) */.use_hse_bypass(8.mhz()).enable_css().freeze();
```

* f0xx
```rust
let rcc = dp.RCC.configure().hse(HSEClock::new(8.mhz(), HSEBypassMode::Bypassed)).freeze(&mut dp.FLASH);
```

* f7xx
```rust
let clocks = rcc.cfgr.hse(HSEClock::new(25.mhz(), HSEClockMode::Bypass)).freeze();
```

* g0xx, g4xx
```rust
let rcc = rcc.freeze(Config::pll().pll_cfg(PllConfig { max: PLLSrc::HSE_BYPASS(8.mhz()), ..Default::default() }));
```

* h7xx
```rust
let ccdr = rcc.use_hse(25.mhz()).bypass_hse().freeze(pwrcfg, &dp.SYSCFG);
```

* l4xx
```rust
let clocks = rcc.cfgr.hse(8.mhz(), CrystalBypass::Disable, ClockSecuritySystem::Disable).freeze(&mut flash.acr, &mut pwr);
```

### DIFF
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -87,7 +87,7 @@ impl AHB {
 /// ```
 /// let dp = pac::Peripherals::take().unwrap();
 /// let rcc = dp.RCC.constrain();
-/// use_ahb(&mut rcc.apb1)
+/// use_apb1(&mut rcc.apb1)
 /// ```
 pub struct APB1 {
     _0: (),
@@ -112,7 +112,7 @@ impl APB1 {
 /// ```
 /// let dp = pac::Peripherals::take().unwrap();
 /// let rcc = dp.RCC.constrain();
-/// use_ahb(&mut rcc.apb2)
+/// use_apb2(&mut rcc.apb2)
 /// ```
 pub struct APB2 {
     _0: (),
@@ -212,7 +212,7 @@ impl BDCR {
 /// ```
 /// let dp = pac::Peripherals::take().unwrap();
 /// let rcc = dp.RCC.constrain();
-/// use_ahb(&mut rcc.cfgr)
+/// use_cfgr(&mut rcc.cfgr)
 /// ```
 pub struct CFGR {
     hse: Option<u32>,
@@ -376,9 +376,10 @@ impl CFGR {
             }
 
             // PLL_MUL maximal value is 16
-            assert!(divisor <= 16);
-            // PRE_DIV maximal value is 16
             assert!(multiplier <= 16);
+
+            // PRE_DIV maximal value is 16
+            assert!(divisor <= 16);
 
             (multiplier, Some(divisor))
         }
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -283,31 +283,28 @@ fn into_pre_div(div: u8) -> cfgr2::PREDIV_A {
 }
 
 impl CFGR {
-    /// Enable HSE (external clock) in crystal mode.
-    /// Uses external oscillator instead of HSI (internal RC oscillator) as the clock source.
+    /// Uses HSE (external oscillator) instead of HSI (internal RC oscillator) as the clock source.
     /// Will result in a hang if an external oscillator is not connected or it fails to start.
     pub fn use_hse<F>(mut self, freq: F) -> Self
     where
         F: Into<Hertz>,
     {
         self.hse = Some(freq.into().0);
-        self.hse_bypass = false;
         self
     }
 
-    /// Enable HSE (external clock) in bypass mode.
-    /// Uses user provided clock instead of HSI (internal RC oscillator) as the clock source.
-    /// Will result in a hang if an external clock source is not connected.
-    pub fn use_hse_bypass<F>(mut self, freq: F) -> Self
-    where
-        F: Into<Hertz>,
-    {
-        self.hse = Some(freq.into().0);
+    /// Enable HSE bypass.
+    /// Uses user provided clock signal instead of an external oscillator.
+    /// OSC_OUT pin is free and can be used as GPIO.
+    /// No effect if HSE is not enabled.
+    pub fn bypass_hse(mut self) -> Self {
         self.hse_bypass = true;
         self
     }
 
     /// Enable CSS (Clock Security System).
+    /// System clock is automatically switched to HSI and an interrupt (CSSI) is generated
+    /// when HSE clock failure is detected.
     /// No effect if HSE is not enabled.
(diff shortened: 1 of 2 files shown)


src/rcc.rs
```diff
@@ -286,13 +283,34 @@ fn into_pre_div(div: u8) -> cfgr2::PREDIV_A {
...
     {
         self.hse = Some(freq.into().0);
+        self.hse_bypass = false;
+        self
+    }
+
+    /// Enable HSE (external clock) in bypass mode.
+    /// Uses user provided clock instead of HSI (internal RC oscillator) as the clock source.
+    /// Will result in a hang if an external clock source is not connected.
+    pub fn use_hse_bypass<F>(mut self, freq: F) -> Self
```
Sh3Rm4n, line 301: Maybe make it more clear, what the purpose of the bypass mode is?
I had to google to understand, what hse_bypass is for. 

> HSE crystal oscillator bypassed with external clock

is used for the documentation of the bypass flag. 
Maybe:

```suggestion
    /// Enable HSE (external clock) in bypass mode.
    /// Uses user provided clock instead of a crystal oscillator. 
    /// Will result in a hang if an external clock source is not connected.
    pub fn use_hse_bypass<F>(mut self, freq: F) -> Self
```

Which makes it more clear, that with bypass mode enabled, the microcontroller won't drive the XTAL pins anymore, as it would do, if an external crystal would be present and (bypass mode inactive).

Sh3Rm4n (changes requested):
Hey thanks for stepping forward and trying to fix this issue.

I'm not sure if we should use a `use_hse_crystal()` and a `use_hse_bypass()`, which both accept a frequency. These options are mutually exclusive.

I would prefer `use_hse().bypass_crystal()` maybe? Or something similar.

Thank you for the comparison to the different stm32 APIs!
As an ad hoc solution I would prefer the one from 'h7xx'.

---

A bit off-topic for this PR:

But I think an API rework has to be done for the `rcc pll` configuration.
For example, wrong configurations are resulting into panics (which I'm trying to address in https://github.com/stm32-rs/stm32f3xx-hal/pull/153).

But this is not an ideal solution. I would also like to prefer an API, which lets us easily create / calculate the PLL configuration in a separate function, which potentially could be executed in a `const` context (when `const fn` has enough stable features available).

---

In conclusion:

For this PR, the `h7xx`-solution is sufficient.

But for a major API rework I would prefer the `g0xx, g4xx` approach, even though it looks ugly. Maybe the 'f7xx' approach as an alternative, which doesn't seem as flexible, but looks cleaner.
I'll have to investigate.

Thanks for the documentation improvements! They are much appreciated, as they are easily overlooked.

David-OConnor:
I don't like the word *crystal*, since there are alternative oscillators like MEMS.

David-OConnor:
See https://github.com/stm32-rs/stm32f3xx-hal/pull/159

Sh3Rm4n:
>I don't like the word crystal, since there are alternative oscillators like MEMS.

What alternatvie would you suggest to differentiate between oscillators, which have to be driven and active clocks (for which the bypass mode is used)?

David-OConnor:
`bypass_hse`

Piroro-hs:
Updated to h7xx style as a temporal solution.

I agree that more explicit APIs (such as g0xx, g4xx approach and #159) would be better.

Sh3Rm4n (approved):
Will merge after CI is all green (rustfmt)

Piroro-hs:
Fixed.

However, IDK why rustfmt is complaining about
https://github.com/stm32-rs/stm32f3xx-hal/blob/d10a12240bca4d5faf67fa4d02ee62667113d6a2/src/rcc.rs#L604-L607
while
https://github.com/stm32-rs/stm32f3xx-hal/blob/92d9357805b1bea0ef9e674f60db1bcff8d1f8df/src/i2c.rs#L257-L264
is allowed.

strom-und-spiele:
> https://github.com/stm32-rs/stm32f3xx-hal/blob/92d9357805b1bea0ef9e674f60db1bcff8d1f8df/src/i2c.rs#L257-L264
> is "allowed".
because it is inside a macro and rustfmt does not seem to care.

I'd suggest using an auto linter and just let it do its thing so one gets used to the default style. 
And if you find yourself changing the same line over and over again - fighting the linter for readability - there is (as a last resort) 
`#[rustfmt::skip]`

Sh3Rm4n:
`rustfmt` is definitly not perfect. There are options, we could set in a config file. I have'nt played around with it yet. 
More importantly is a consitent style.

PR #158, opened 2020-10-25
Clocks

David-OConnor:
It seems like the current system fails silently, and gives the user too many degrees of freedom. Eg the illusion you can set whatever speeds you want. I think the STM32Cube IDE visual clock config is a good guide for what configs are valid.

Suggestion: Instead of setting speeds, set the various scalers. This alone reduces the degrees of freedom. Have a helper method to show what speeds these output. Calculate the speeds, and validate before setting. This deals with errors in 2 ways: Prevents requesting speeds that can't be set using the scalers and 2: Fails explicitly if the resulting speeds are out of range.

Proof of concept. Not ready to be merged, but I think the structure is sound. Missing features, needs QC, and needs to set up some side effects like flash.


David-OConnor:
Closed due to wrong base

PR #169, opened 2020-12-02
Seal "private" traits 🦭

Sh3Rm4n:
Applying the "[Sealed Traits](https://rust-lang.github.io/api-guidelines/future-proofing.html)" pattern 
to make the mentioned trait really private. 

This reduces the amount of unsafe calls, which in this case are not related to any memory safety, 
and resolves the mentioned `FIXME`s in the code. 

I'm not happy about the code-dublication though. Especially in src/spi.rs there are lots of it. 
And I imagine, it won't be fun to adjust the code, when configuration errors are detected. 

Maybe this should be solved with a macro. But as modules can't be split apart, I've not come up with a good solution, yet.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,7 +64,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.50.0
+          toolchain: 1.51.0
           target: thumbv7em-none-eabihf
           override: true
           profile: minimal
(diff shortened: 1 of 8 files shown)


Sh3Rm4n:
Using macros now. .

Not quite satisfied either. Because now I had to sacrifice the convenience from `cfg_if!` ...

Piroro-hs:
This could be improved after #189 landed.

```rust
// lib.rs
mod private {
    pub trait Sealed {}
}

// gpio.rs
use crate::private::Sealed;
impl<Gpio, Index, Mode> Sealed for Pin<Gpio, Index, Mode> {}

// serial.rs (for example)
use crate::private::Sealed;
pub trait TxPin<USART>: Sealed {}
impl<Otype> TxPin<USART1> for gpioa::PA9<AF7<Otype>> {}

```

PR #178, opened 2020-12-09
Can refactor

Sh3Rm4n:
- [ ] Use echo as example or at least as test


Piroro-hs:
How about using [bxcan](https://github.com/stm32-rs/bxcan) crate?

Sh3Rm4n:
This will be pretty much superseded by #207

PR #186, opened 2020-12-20
Small fixes for CAN

zklapow:
I know there is an open issue for refactoring the CAN impl (#178) but figured I would patch the existing impl slightly in the meantime.

This fixes an issue where weren't actually transmitting the correct length of data, as well as adding options to allow users to configure the baud rate and enable loopback mode.

### DIFF
--- a/src/can.rs
+++ b/src/can.rs
@@ -578,7 +578,7 @@ impl CanFrame {
     ///
     /// This function will panic if length of `data` is greater than `8`
     pub fn new_data(id: CanId, data: &[u8]) -> CanFrame {
-        crate::assert!((0..8).contains(&data.len()));
+        crate::assert!((0..=8).contains(&data.len()));
 
         let mut frame = Self {
             id,
--- a/src/can.rs
+++ b/src/can.rs
@@ -30,38 +30,44 @@ const MAX_EXTENDED_ID: u32 = 0x1FFF_FFFF;
 ///
 /// Use `CanOpts::default()` to get 250kbps at 32mhz system clock
 pub struct CanOpts {
-    pub brp: u16,
-    pub sjw: u8,
-    pub ts1: u8,
-    pub ts2: u8,
-    pub lbkm: LBKM_A,
+    brp: u16,
+    sjw: u8,
+    ts1: u8,
+    ts2: u8,
+    lbkm: LBKM_A,
 }
 
 impl CanOpts {
+    /// Create new `CanOpts` using the default settings from `CanOpts::default()` to get 250kbps at 32mhz system clock.
     pub fn new() -> CanOpts {
         CanOpts::default()
     }
 
+    /// Set the Baud Rate Prescaler. See  http://www.bittiming.can-wiki.info/#bxCAN for generating the timing parameters for different baud rates and clocks.
     pub fn brp(mut self, brp: u16) -> Self {
         self.brp = brp;
         self
     }
 
+    /// Set the Resynchronisation Jump Width. See  http://www.bittiming.can-wiki.info/#bxCAN for generating the timing parameters for different baud rates and clocks.
     pub fn sjw(mut self, sjw: u8) -> Self {
         self.sjw = sjw;
         self
     }
 
+    /// Set Time Segment One. See  http://www.bittiming.can-wiki.info/#bxCAN for generating the timing parameters for different baud rates and clocks.
     pub fn ts1(mut self, ts1: u8) -> Self {
         self.ts1 = ts1;
         self
     }
 
+    /// Set Time Segment Two. See  http://www.bittiming.can-wiki.info/#bxCAN for generating the timing parameters for different baud rates and clocks.
     pub fn ts2(mut self, ts2: u8) -> Self {
         self.ts2 = ts2;
         self
     }
 
+    /// Enable or disable loopback mode on the CAN device. This is useful for debugging.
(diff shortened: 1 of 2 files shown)


Sh3Rm4n:
Don't worry about #178 too much. This is still very WIP and for now only contains added TODO comments. More of a reminder for myself. Maybe I should close that PR until it contains some useful changes.

src/can.rs
```diff
@@ -24,6 +24,36 @@ use core::sync::atomic::{AtomicU8, Ordering};
 const EXID_MASK: u32 = 0b1_1111_1111_1100_0000_0000_0000_0000;
 const MAX_EXTENDED_ID: u32 = 0x1FFF_FFFF;
 
+/// Options the CAN bus. This is primarily used to set bus timings, but also controls options like enabling loopback or silent mode for debugging.
+/// See  http://www.bittiming.can-wiki.info/#bxCAN for generating the timing parameters for different baud rates and clocks.
+///
+/// Use `CanBitRateOpts::default()` to get 250kbps at 32mhz system clock
```
Sh3Rm4n, line 30:
```suggestion
/// Use `CanOpts::default()` to get 250kbps at 32mhz system clock
```

src/can.rs
```diff
@@ -24,6 +24,36 @@ use core::sync::atomic::{AtomicU8, Ordering};
...
+pub struct CanOpts {
+    pub brp: u16,
+    pub sjw: u8,
+    pub ts1: u8,
+    pub ts2: u8,
+    pub lbkm: bool,
+}
```
Sh3Rm4n, line 37: I'm not a big fan of this approach. 

Configuring via integer values does not really guide a user of this crate towards a correct and comprehensible configuration of this peripheral, and it does lead to having many magic values in the code. 

Maybe use enums, possibly leveraging the already existing enum types provided by [stm32-rs] use the [builder-pattern](https://github.com/rust-unofficial/patterns/blob/master/patterns/builder.md)? 

[stm32-rs]: https://docs.rs/stm32f3/0.12.1/stm32f3/stm32f303/can/btr/index.html

Sh3Rm4n (changes requested):

zklapow:
updated this to use a builder + the source enums where possible.

zklapow:
(also didnt realize the clippy lints were failing, fixed now)

src/can.rs
```diff
@@ -20,10 +20,72 @@ use crate::stm32;
...
+    }
+
+    /// Set Time Segment Two. See  http://www.bittiming.can-wiki.info/#bxCAN for generating the timing parameters for different baud rates and clocks.
+    pub fn ts2(mut self, ts2: u8) -> Self {
+        self.ts2 = ts2;
+        self
+    }
+
+    /// Enable or disable loopback mode on the CAN device. This is useful for debugging.
+    pub fn lbkm(mut self, lbkm: LBKM_A) -> Self {
```
Sh3Rm4n, line 71: Maybe reexport the `LBKM_A` type, so that the user does not have to dig through the stm32::can abstractions to find the correct enum.

src/can.rs
```diff
@@ -312,6 +360,15 @@ impl Can {
             _tx: tx,
         }
     }
+    /// Initialize the CAN Peripheral
```
Sh3Rm4n, line 363: Minor nit: In this comment you could mention, that the options are implicitly created by `CanOpts::default()`, as I find this information particularly useful to know here. 

Maybe something like  

```suggestion
    /// Initialize the CAN Peripheral
    /// 
    /// The CAN peripheral is configured according to `CanOpts::default()`.
    /// If you want to explicitly set the configuration, use `Can::new_with_opts()` instead.
```

Sh3Rm4n (changes requested):
Sorry for the long delay on my side again. I've to figure out to devote some time regularly to this project. The last active maintenance efforts of myself are reduced to short bursts. 

Some very minor nits, the rest looks ok to me. As `Can::new()` has the same API as before, this is not API break. Changelog entries are always welcome, but in this case not required.

In the long run, we should think about using the https://github.com/stm32-rs/bxcan types, 
and in the longer run, see if we can copy / merge some ideas from https://github.com/stm32-rs/stm32f1xx-hal/pull/293. 

But this takes time. As of now, I'm really happy, that you are doing this work on the CAN peripheral right now! :)

zklapow:
no prob @Sh3Rm4n I am in the same boat myself. I find occasional bouts of time to work on this, but much less than I would prefer!

I will make some of those changes later today and update the changelog as well.

I totally agree about using `bxCan`, I was looking for something like that to exist when I originally started work on this! I have been meaning to give that a go, but have been hesitant to mess with it because I have been using the existing CAN stuff for a motor driver I am building and didn't want to mess with it. I might give it a go soon though.

Sh3Rm4n:
> I will make some of those changes later today and update the changelog as well.

Nice :+1: 

> I totally agree about using bxCan, I was looking for something like that to exist when I originally started work on this! 

Yeah, I know that your original work on CAN predates the `bxCAN` crate. It's nice to see the eco-system improving :)

> I have been meaning to give that a go, but have been hesitant to mess with it because I have been using the existing CAN stuff for a motor driver I am building and didn't want to mess with it.

No pressure. Maybe some else has time to pick this up and you can focus on your motor driver! :)

Sh3Rm4n (approved):
LGTM

Do want to add anything or can I merge it?

zklapow:
Should be good to merge!

Issue #125, opened 2020-07-26
Collapse Input<OpenDrain / PullUp / PullDown) into Input


mattiasgronlund:
### Rethink Pin Type States in Regard of Pull Up/Down 

Currently Input<> has three different states (Floating, PullUp, PullDown) which expose the pull-up and pull-down settings.
But Output / AF has no such distinction, instead there is a method internal_pull_up(bool) available for AF and Output<OpenDrain>, which make it possible to switch between floating of pull-up.
I see a couple of problems with this model:
1. it is not consistent
2. it limits the available settings more than the MCU really does, which might be a problem in some situations.
3. I would call the current implementation as being buggy, as it currently allows one to fool it. You can move a pin into Input<PullDown> then move it to Output<OpenDrain> or even Output<PushPull> and as neither of the later moves resets the pull up/down states you enter a state which should have been "blocked".

My suggestion is that we collapse the Input<*> states into just Input.

One other solution would be to actually add more type states, but I think that would be a bad idea as the pull up/down is more of a hardware choice to have in the MCU and not external on the board.

Sh3Rm4n:
[`Floating, PullUp, PullDown`](https://github.com/stm32-rs/stm32f3xx-hal/blob/master/src/gpio.rs#L30-L35) for Input is, what [`OpenDrain, PushPull`](https://github.com/stm32-rs/stm32f3xx-hal/blob/master/src/gpio.rs#L42-L45) is for Output. 

> One other solution would be to actually add more type states, but I think that would be a bad idea as the pull up/down is more of a hardware choice to have in the MCU and not external on the board.

The stm32f4xx-hal somewhat recently added a [`OpenDrain`](https://github.com/stm32-rs/stm32f4xx-hal/pull/122/files) implementation for AFs. So this seems like being the trend for the other HALs, at least for Output / AF pins. 

> My suggestion is that we collapse the Input<*> states into just Input.

I see that this is a viable solution. I can't really think of a situation, where it is needed to encode the pull state in a typestate and as you said, an Input pin can be `Floating` internally, but still a pull-up resistors is attached externally. For consistency, there should be a method to transfer the Input from `Input<Floating>` to `Input<Pullup>` without activating the internal pull-up, which seems strange. 

Point **3.** seems rather important though. I would classify this as an implementation bug we have to fix.

Sh3Rm4n:
[...]

Issue #112, opened 2020-07-14
Interrupt support


David-OConnor:
I'd like to see this library include high-level code to handle interrupts, and am curious how y'all would like to approach this. @jkboyce  , who posted [this issue](https://github.com/stm32-rs/stm32f3xx-hal/issues/83) has been immensely helpful, as have several people on the `stm32` Matrix chat, especially `adamgreig`.

I've been working off a standalone module. Here are possible approaches:
1 - Achieve API and implementation parity with other stm32 hal crates. I haven't done my research on this, although understanding how this API is set up isn't trivial due to heavy use of macros and traits. API parity has obvious advantages - most notably the main HAL virtue of transparent switching between MCUs, so this is the ideal solution.

2 - Add a standalone module to this crate using hand-written, repetitive fns

3 - Same as above, but using macros to make it cleaner to maintain. I'm not sure how to do this, as my trivial attempts have failed. Might need to use the [paste crate](https://crates.io/crates/paste) or proc macros.

4 - As a standalone module released on `crates.io`. This is the backup approach. I would never have been able to get interrupts working in my crate without the help of the two people listed at the top of this post, and want this to be a solved problem going forward.

Here are the basics for GPIO interrupts, using option 2. It works, but suffers from DRY. The remaining interrupt functionality (The 4 types of RTC interrupts, I2c interrupts etc) should be added as well, either in the initial PR, or later.

```rust
/// Reference the STM32F303 reference manual, section 14.2.5 for a `Functional description`
/// of the steps we accomplish here.
/// 0 - 15.
pub fn setup_line(
    exti: &mut EXTI,
    line: u8,
    task: Task,
    edge: Edge,
) {
    // Configure the Trigger Selection bits of the Interrupt line (EXTI_RTSR and EXTI_FTSR)
    let rise_trigger = match edge {
        Edge::Rising => {
            // configure EXTI line to trigger on rising edge, disable trigger on falling edge.
            true
        }
        Edge::Falling => {
            // configure EXTI line to trigger on falling edge, disable trigger on rising edge
           false
        }
    };

    match line {
        0 => {
            // Configure the corresponding mask bit in the EXTI_IMR register.
            exti.imr1.modify(|_, w| w.mr0().unmasked());
            exti.rtsr1.modify(|_, w| w.tr0().bit(rise_trigger));
            exti.ftsr1.modify(|_, w| w.tr0().bit(!rise_trigger));

[...]

PR #189, opened 2020-12-24
GPIO refactor (Simplify macro, reduce type states, and support interrupt)

### LINKED ISSUE #83: External interrupts (example, question)
I wrote a simple example of GPIO-driven interrupt handling, available [here.](https://github.com/jkboyce/stm32f3xx-hal/blob/master/examples/gpio_interrupt.rs) Compile with `cargo build --example gpio_interrupt --features "rt stm32f303xc"` On the F3 Discovery board it toggles an led when you push the user button.

I'm curious if anyone's working on–or has thoughts about–adding better support for interrupts. The [`gpio.rs`](https://github.com/stm32-rs/stm32f4xx-hal/blob/master/src/gpio.rs) in `stm32f4xx-hal` has some nice methods based on the `ExtiPin` trait, making it easy to set up interrupts without direct register access. For example:

```rust
let gpioc = dp.GPIOC.split();
let mut board_btn = gpioc.pc13.into_floating_input();
board_btn.make_interrupt_source(&mut dp.SYSCFG);
board_btn.enable_interrupt(&mut dp.EXTI);
board_btn.trigger_on_edge(&mut dp.EXTI, Edge::RISING);
```

It would be nice to get the f3/f4 HALs into closer alignment but I wanted to gauge others' thoughts before jumping in.

Piroro-hs:
* Simplify macro
  `gpio.rs` uses more generics than macro now
* Collapse Input type arguments
  An attempt to address #125
* Support interrupt
  Close #83 and #112

### DIFF
--- a/examples/gpio_interrupts.rs
+++ b/examples/gpio_interrupts.rs
@@ -12,7 +12,7 @@ use cortex_m_rt::entry;
 use hal::gpio::{gpioa, gpioe, Edge, Input, Output, PushPull};
 use hal::interrupt;
 use hal::pac;
-use hal::pac::{Interrupt, NVIC};
+use hal::pac::NVIC;
 use hal::prelude::*;
 
 type LedPin = gpioe::PE9<Output<PushPull>>;
@@ -21,7 +21,7 @@ static LED: Mutex<RefCell<Option<LedPin>>> = Mutex::new(RefCell::new(None));
 type ButtonPin = gpioa::PA0<Input>;
 static BUTTON: Mutex<RefCell<Option<ButtonPin>>> = Mutex::new(RefCell::new(None));
 
-// When the user button is pressed. The north LED with toggle.
+// When the user button is pressed. The north LED will toggle.
 #[entry]
 fn main() -> ! {
     // Getting access to registers we will need for configuration.
@@ -48,10 +48,12 @@ fn main() -> ! {
     user_button.make_interrupt_source(&mut syscfg);
     user_button.trigger_on_edge(&mut exti, Edge::Rising);
     user_button.enable_interrupt(&mut exti);
+    let interrupt_num = user_button.nvic(); // hal::pac::Interrupt::EXTI0
+
     // Moving ownership to the global BUTTON so we can clear the interrupt pending bit.
     cortex_m::interrupt::free(|cs| *BUTTON.borrow(cs).borrow_mut() = Some(user_button));
 
-    unsafe { NVIC::unmask(Interrupt::EXTI0) };
+    unsafe { NVIC::unmask(interrupt_num) };
 
     loop {
         asm::wfi();
--- a/examples/gpio_interrupts.rs
+++ b/examples/gpio_interrupts.rs
@@ -1,0 +1,86 @@
+#![no_main]
+#![no_std]
+
+use panic_semihosting as _;
+
+use stm32f3xx_hal as hal;
+
+use core::cell::RefCell;
+use cortex_m::asm;
+use cortex_m::interrupt::Mutex;
+use cortex_m_rt::entry;
+use hal::gpio::{gpioa, gpioe, Edge, Input, Output, PushPull};
+use hal::interrupt;
+use hal::pac;
+use hal::pac::{Interrupt, NVIC};
+use hal::prelude::*;
+
+type LedPin = gpioe::PE9<Output<PushPull>>;
+static LED: Mutex<RefCell<Option<LedPin>>> = Mutex::new(RefCell::new(None));
+
+type ButtonPin = gpioa::PA0<Input>;
+static BUTTON: Mutex<RefCell<Option<ButtonPin>>> = Mutex::new(RefCell::new(None));
+
+// When the user button is pressed. The north LED with toggle.
+#[entry]
+fn main() -> ! {
+    // Getting access to registers we will need for configuration.
+    let device_peripherals = pac::Peripherals::take().unwrap();
+    let mut rcc = device_peripherals.RCC.constrain();
+    let mut syscfg = device_peripherals.SYSCFG.constrain(&mut rcc.apb2);
+    let mut exti = device_peripherals.EXTI;
+    let mut gpioe = device_peripherals.GPIOE.split(&mut rcc.ahb);
+    let mut gpioa = device_peripherals.GPIOA.split(&mut rcc.ahb);
+
+    let mut led = gpioe
+        .pe9
+        .into_push_pull_output(&mut gpioe.moder, &mut gpioe.otyper);
+    // Turn the led on so we know the configuration step occurred.
+    led.toggle().expect("unable to toggle led in configuration");
+
+    // Move the ownership of the led to the global LED
+    cortex_m::interrupt::free(|cs| *LED.borrow(cs).borrow_mut() = Some(led));
+
+    // Configuring the user button to trigger an interrupt when the button is pressed.
+    let mut user_button = gpioa
+        .pa0
+        .into_pull_down_input(&mut gpioa.moder, &mut gpioa.pupdr);
+    user_button.make_interrupt_source(&mut syscfg);
+    user_button.trigger_on_edge(&mut exti, Edge::Rising);
+    user_button.enable_interrupt(&mut exti);
+    // Moving ownership to the global BUTTON so we can clear the interrupt pending bit.
+    cortex_m::interrupt::free(|cs| *BUTTON.borrow(cs).borrow_mut() = Some(user_button));
+
+    unsafe { NVIC::unmask(Interrupt::EXTI0) };
+
+    loop {
+        asm::wfi();
+    }
+}
+
+// Button Pressed interrupt.
+// The exti# maps to the pin number that is being used as an external interrupt.
+// See page 295 of the stm32f303 reference manual for proof:
+// http://www.st.com/resource/en/reference_manual/dm00043574.pdf
+//
+// This may be called more than once per button press from the user since the button may not be debounced.
+#[interrupt]
+fn EXTI0() {
+    cortex_m::interrupt::free(|cs| {
+        // Toggle the LED
+        LED.borrow(cs)
+            .borrow_mut()
+            .as_mut()
+            .unwrap()
+            .toggle()
+            .unwrap();
+
+        // Clear the interrupt pending bit so we don't infinitely call this routine
+        BUTTON
+            .borrow(cs)
+            .borrow_mut()
+            .as_mut()
+            .unwrap()
+            .clear_interrupt_pending_bit();
+    })
+}
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -118,6 +118,9 @@ where
 /// Marker trait for readable pin modes
 pub trait Readable {}
 
+/// Marker trait for PullUppable pin modes
+pub trait PullUppable {} // TODO: better naming
+
 /// Runtime defined GPIO port (type state)
 pub struct Gpiox {
     ptr: *const dyn GpioRegExt,
@@ -155,19 +158,21 @@ pub struct PullDown;
 /// Output mode (type state)
 pub struct Output<MODE>(PhantomData<MODE>);
 
-/// Push pull output (type state)
+/// Push-pull output (type state)
 pub struct PushPull;
-/// Open drain output (type state)
+/// Open-drain output (type state)
 pub struct OpenDrain;
 
 /// Alternate function (type state)
-pub struct Alternate<AF>(PhantomData<AF>);
+pub struct Alternate<AF, MODE>(PhantomData<AF>, PhantomData<MODE>);
 
 /// Analog mode (type state)
 pub struct Analog;
 
 impl<MODE> Readable for Input<MODE> {}
 impl Readable for Output<OpenDrain> {}
+impl PullUppable for Output<OpenDrain> {}
+impl<AF> PullUppable for Alternate<AF, OpenDrain> {}
 
 /// Generic pin
 pub struct Pin<GPIO, INDEX, MODE> {
@@ -306,10 +311,11 @@ where
     }
 }
 
-impl<GPIO, INDEX> Pin<GPIO, INDEX, Output<OpenDrain>>
+impl<GPIO, INDEX, MODE> Pin<GPIO, INDEX, MODE>
 where
     GPIO: GpioStatic,
     INDEX: Index,
+    MODE: PullUppable,
 {
     /// Enables / disables the internal pull up
     pub fn internal_pull_up(&mut self, pupdr: &mut GPIO::PUPDR, on: bool) {
@@ -385,12 +391,7 @@ where
 }
 
 macro_rules! af {
-    ($i:literal, $Ui:ty, $AFi:ty, $IntoAfi:ident, $into_afi:ident) => {
-        paste::paste! {
-            #[doc = "Alternate function " $i " (type state)"]
-            pub type $AFi = Alternate<$Ui>;
-        }
-
+    ($i:literal, $Ui:ty, $AFi:ident, $IntoAfi:ident, $into_afi_push_pull:ident, $into_afi_open_drain:ident) => {
         paste::paste! {
             #[doc = "Marker trait for pins with alternate function " $i " mapping"]
             pub trait $IntoAfi {
@@ -399,19 +400,39 @@ macro_rules! af {
             }
         }
 
+        paste::paste! {
+            #[doc = "Alternate function " $i " (type state)"]
+            pub type $AFi<MODE> = Alternate<$Ui, MODE>;
+        }
+
         impl<GPIO, INDEX, MODE> Pin<GPIO, INDEX, MODE>
         where
             Self: $IntoAfi,
             GPIO: GpioStatic,
             INDEX: Index,
         {
-            /// Configures the pin to operate as an alternate function pin
-            pub fn $into_afi(
+            /// Configures the pin to operate as an alternate function push-pull output pin
+            pub fn $into_afi_push_pull(
                 self,
                 moder: &mut GPIO::MODER,
+                otyper: &mut GPIO::OTYPER,
                 afr: &mut <Self as $IntoAfi>::AFR,
-            ) -> Pin<GPIO, INDEX, $AFi> {
+            ) -> Pin<GPIO, INDEX, $AFi<PushPull>> {
                 moder.alternate(self.index.index());
+                otyper.push_pull(self.index.index());
+                afr.afx(self.index.index(), $i);
+                self.into_mode()
+            }
+
+            /// Configures the pin to operate as an alternate function open-drain output pin
+            pub fn $into_afi_open_drain(
+                self,
+                moder: &mut GPIO::MODER,
+                otyper: &mut GPIO::OTYPER,
+                afr: &mut <Self as $IntoAfi>::AFR,
+            ) -> Pin<GPIO, INDEX, $AFi<OpenDrain>> {
+                moder.alternate(self.index.index());
+                otyper.open_drain(self.index.index());
                 afr.afx(self.index.index(), $i);
                 self.into_mode()
             }
@@ -421,7 +442,7 @@ macro_rules! af {
     ([$($i:literal),+ $(,)?]) => {
         paste::paste! {
             $(
-                af!($i, [<U $i>], [<AF $i>], [<IntoAf $i>],  [<into_af $i>]);
+                af!($i, [<U $i>], [<AF $i>], [<IntoAf $i>],  [<into_af $i _push_pull>],  [<into_af $i _open_drain>]);
             )+
         }
     };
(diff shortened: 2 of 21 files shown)


Sh3Rm4n:
Super interesting attempt. I'll take a closer look in the next few days. But this approach seems promising!

src/gpio.rs
```diff
@@ -50,1205 +46,1289 @@ pub trait GpioExt {
     fn split(self, ahb: &mut AHB) -> Self::Parts;
 }
 
-/// Input mode (type state)
-pub struct Input<MODE> {
-    _mode: PhantomData<MODE>,
+/// GPIO Register interface traits private to this module
```
Sh3Rm4n, line 49: Also known as sealed traits :)

src/gpio.rs
```diff
@@ -50,1205 +46,1289 @@ pub trait GpioExt {
...
+impl Index for Ux {
+    fn index(&self) -> u8 {
+        self.0
+    }
+}
 
-/// Alternate function 9 (type state)
-pub struct AF9;
+/// Input mode (type state)
+pub struct Input;
```
Sh3Rm4n, line 160: Maybe add a documentation notice, that the Input type is not encoded in a type state anymore?
The user has to know, that it is their responsibility to set the correct input type and that this 
is due issues explained in https://github.com/stm32-rs/stm32f3xx-hal/issues/125. 

The change notice belonging to the changelog, the overall behavior should be written down here.

src/gpio.rs
```diff
@@ -50,1205 +46,1289 @@ pub trait GpioExt {
...
+/// Output mode (type state)
+pub struct Output<OTYPE>(PhantomData<OTYPE>);
```
Sh3Rm4n, line 162: The elements of tuples are always public, iirc. Is this intended or a problem?

src/gpio.rs
```diff
@@ -50,1205 +46,1289 @@ pub trait GpioExt {
...
+/// Alternate function (type state)
+pub struct Alternate<AF, OTYPE>(PhantomData<AF>, PhantomData<OTYPE>);
```
Sh3Rm4n, line 164: I've noticed many changes, where `Alternate<PushPull>` is now enforced. 
Is this really needed for every case? [`stm32f4xx-hal`](https://github.com/stm32-rs/stm32f4xx-hal/blob/dadf2192e6ade2652de3b02754c3606532dd624d/src/gpio.rs#L34-L42) does only differentiate between `Alternate` and `Alternate<OpenDrain>` / `AlternateOD`. 

Is there a use case for a neutral `Alternate` Type State, where it does not matter, which Output State is used?
Point 3. in https://github.com/stm32-rs/stm32f3xx-hal/issues/125#issue-665855804 might be a problem, right? 

What is your thought about this?

Sh3Rm4n (commented):
Some thoughts which came into my mind. 

I've only read a small part of `src.gpio.rs` so far. Will look into this in more detail later.

src/gpio.rs
Piroro-hs: I follow the [stm32f1xx-hal](https://github.com/stm32-rs/stm32f1xx-hal/blob/ca01418e356d2879628dd86cc4c63597d85c1665/src/gpio.rs#L484-L530) as it feels more consistent.

Some peripherals should accept `Alternate<MODE>` instead of `Alternate<PushPull>`, maybe?

src/gpio.rs
Piroro-hs: Tuple elements are [treated like normal struct fields](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=966728d4d6f3661263b58275910d4483), so this is ok 😃

src/gpio.rs
Piroro-hs: Yeah, documentation update is needed.

However, I currently think about re-adding `into_input_xxx()` methods which implemented like
```rust
pub fn into_input_floating(self, moder: &mut GPIO::MODER, pupdr: &mut GPIO::PUPDR) -> Pin<GPIO, INDEX, Input> {
    pupdr.floating(self.index.index());
    self.into_input(moder)
}
```

David-OConnor:
Could you please add support for GPIO interrupts? I'm actually modding this now to do so. I made a standalone module a while back for this (In my PR), but am switching it to GPIO to be consistent with L4 and F4 HALs.

Piroro-hs:
GPIO interrupts support added.
Referred to f1xx-hal, f4xx-hal, and #200.

Piroro-hs:
Clippy warnings should be fixed by #202

src/gpio.rs
Piroro-hs: Support for OD pins is added for serial and pwm peripherals.

src/gpio.rs
```diff
@@ -50,1205 +73,1450 @@ pub trait GpioExt {
...
+/// at runtime, and erases the specific compile time type of the GPIO.
+/// It does only matter, that it is a GPIO pin with a specific MODE.
+///
+/// See [examples/gpio_erased.rs] as an example.
+///
+/// [examples/gpio_erased.rs]: https://github.com/stm32-rs/stm32f3xx-hal/blob/v0.6.0/examples/gpio_erased.rs
+pub type PXx<MODE> = Pin<Gpiox, Ux, MODE>;
+
+macro_rules! modify_at {
+    ($xr:expr, $bits:expr, $i:expr, $val:expr) => {
```
seanybaggins, line 263: `pin_num` could be a more informative name than `i`.

seanybaggins: Perhaps adding required tokens to the macro could help improve readability when the macro is invoked.
Ex.
```rust
macro_rules! modify_at {
    (register: $xr:expr, some_name_1: $bits:expr, extri_num: $i:expr, some_name: $val:expr) => {
        ...
    }    
}
```

src/gpio.rs
```diff
@@ -50,1205 +73,1450 @@ pub trait GpioExt {
...
+    GPIO: Gpio,
+    INDEX: Index,
+    MODE: Active,
+{
+    /// Make corresponding EXTI line sensitive to this pin
+    pub fn make_interrupt_source(&mut self, syscfg: &mut SysCfg) {
+        let i = self.index.index() % 4;
+        let extigpionr = self.gpio.port_index() as u32;
+        match self.index.index() {
+            0..=3 => unsafe { modify_at!(syscfg.exticr1, 4, i, extigpionr) },
```
seanybaggins, line 514: Consider adding some let expression before the `modify_at` macro that labels what the number 4 is.

src/gpio.rs
```diff
@@ -50,1205 +73,1450 @@ pub trait GpioExt {
...
+            12..=15 => unsafe { modify_at!(syscfg.exticr4, 4, i, extigpionr) },
+            _ => unreachable!(),
+        }
+    }
+
+    /// Generate interrupt on rising edge, falling edge, or both
+    pub fn trigger_on_edge(&mut self, exti: &mut EXTI, edge: Edge) {
+        cfg_if! {
+            // if #[cfg(any(feature = "stm32f373", feature = "stm32f378"))] { // TODO: update after https://github.com/stm32-rs/stm32-rs/pull/496 released
+            if #[cfg(not(any(feature = "stm32f302", feature = "stm32f303", feature = "stm32f334")))] {
```
seanybaggins, line 526: I am not sure this conditional compilation is correct. Looking at the [stm32-rs/STM32F301](https://stm32-rs.github.io/stm32-rs/STM32F301.html#EXTI:RTSR1). It does not fall in this feature category but does have a numbered RTSR.

src/gpio.rs
```diff
@@ -50,1205 +73,1450 @@ pub trait GpioExt {
...
+                let ftsr = &exti.ftsr1;
+            }
+        }
+        let (rise, fall) = match edge {
+            Edge::Rising => (true, false),
+            Edge::Falling => (false, true),
+            Edge::RisingFalling => (true, true),
+        };
+        unsafe {
+            modify_at!(rtsr, 1, self.index.index(), rise as u32);
```
seanybaggins, line 540: Consider adding a let expression that labels what 1 is.

src/gpio.rs
```diff
@@ -50,1205 +73,1450 @@ pub trait GpioExt {
...
+            modify_at!(ftsr, 1, self.index.index(), fall as u32);
+        }
+    }
+
+    /// Enable external interrupts from this pin.
+    pub fn enable_interrupt(&mut self, exti: &mut EXTI) {
+        cfg_if! {
+            // if #[cfg(any(feature = "stm32f373", feature = "stm32f378"))] { // TODO: update after https://github.com/stm32-rs/stm32-rs/pull/496 released
+            if #[cfg(not(any(feature = "stm32f302", feature = "stm32f303", feature = "stm32f334")))] {
```
seanybaggins, line 549: Same comment about correct conditional compilation. https://stm32-rs.github.io/stm32-rs/STM32F301.html#EXTI:IMR1

src/gpio.rs
```diff
@@ -50,1205 +73,1450 @@ pub trait GpioExt {
...
+            }
+        }
+        unsafe { modify_at!(imr, 1, self.index.index(), 1) };
+    }
+
+    /// Disable external interrupts from this pin
+    pub fn disable_interrupt(&mut self, exti: &mut EXTI) {
+        cfg_if! {
+            // if #[cfg(any(feature = "stm32f373", feature = "stm32f378"))] { // TODO: update after https://github.com/stm32-rs/stm32-rs/pull/496 released
+            if #[cfg(not(any(feature = "stm32f302", feature = "stm32f303", feature = "stm32f334")))] {
```
seanybaggins, line 562: Same comment about correct conditional compilation. https://stm32-rs.github.io/stm32-rs/STM32F301.html#EXTI:IMR1

src/gpio.rs
```diff
@@ -50,1205 +73,1450 @@ pub trait GpioExt {
...
+            }
+        }
+        unsafe { modify_at!(imr, 1, self.index.index(), 0) };
+    }
+
+    /// Clear the interrupt pending bit for this pin
+    pub fn clear_interrupt_pending_bit(&mut self) {
+        cfg_if! {
+            // if #[cfg(any(feature = "stm32f373", feature = "stm32f378"))] { // TODO: update after https://github.com/stm32-rs/stm32-rs/pull/496 released
+            if #[cfg(not(any(feature = "stm32f302", feature = "stm32f303", feature = "stm32f334")))] {
```
seanybaggins, line 575: Correct conditional compilation? https://stm32-rs.github.io/stm32-rs/STM32F301.html#EXTI:PR1

src/gpio.rs
```diff
@@ -50,1205 +73,1450 @@ pub trait GpioExt {
...
+            $(
+                af!($i, [<U $i>], [<AF $i>], [<IntoAf $i>],  [<into_af $i _push_pull>],  [<into_af $i _open_drain>]);
+            )+
+        }
+    };
+}
 
-        #[cfg(feature = "unproven")]
-        impl <MODE> toggleable::Default for PXx<Output<MODE>> {}
+af!([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]);
```
seanybaggins, line 656: Wow that is a lot cleaner

src/gpio.rs
```diff
@@ -50,1205 +73,1450 @@ pub trait GpioExt {
...
 
-                        fn set_low(&mut self) -> Result<(), Self::Error> {
-                            // NOTE(unsafe, write) atomic write to a stateless register
-                            unsafe { (*$GPIOX::ptr()).bsrr.write(|w| w.bits(1 << (16 + self.i))) }
-                            Ok(())
-                        }
+                impl Afr for AFRL {
+                    #[inline]
+                    fn afx(&mut self, i: u8, x: u8) {
+                        unsafe { modify_at!((*$GPIOX::ptr()).afrl, 4, i, x as u32); }
```
seanybaggins, line 841: Consider let expression that labels what 4 is.

src/gpio.rs
Piroro-hs: The [stm32-rs/STM32F301](https://stm32-rs.github.io/stm32-rs/STM32F301.html#EXTI:RTSR1) is based on latest nightly version of stm32-rs, while we are using v0.12 for now.

src/gpio.rs
```diff
@@ -50,1205 +73,1450 @@ pub trait GpioExt {
...
-    _mode: PhantomData<MODE>,
+use private::{Afr, GpioRegExt, Moder, Ospeedr, Otyper, Pupdr};
+
+/// Marker trait for GPIO ports
+pub trait Gpio: private::Gpio {}
+
+/// Marker trait for compile time defined GPIO ports
+pub trait GpioStatic: Gpio {
+    /// Associated MODER register
+    type MODER: Moder;
```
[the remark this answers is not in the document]
Piroro-hs: `Upper` is a trait bound for a type alias `UPPER`.

src/gpio.rs
```diff
@@ -50,1205 +73,1450 @@ pub trait GpioExt {
...
-/// Output mode (type state)
-pub struct Output<MODE> {
-    _mode: PhantomData<MODE>,
+use private::{Afr, GpioRegExt, Moder, Ospeedr, Otyper, Pupdr};
+
+/// Marker trait for GPIO ports
+pub trait Gpio: private::Gpio {}
+
+/// Marker trait for compile time defined GPIO ports
+pub trait GpioStatic: Gpio {
```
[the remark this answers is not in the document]
Piroro-hs: Yup. https://doc.rust-lang.org/rust-by-example/trait/supertraits.html

Sh3Rm4n:
> I'll take a closer look in the next few days. But this approach seems promising!

Next few months rather. I have this on my TODO list. I'm hopeful that I find time this week :)

src/gpio.rs
Sh3Rm4n: The `modify_at` macro could need some explanation, but 4 is essentially the `bitwidth` of the field in the 32 bit register, which is changed by this invocation.

src/gpio.rs
```diff
@@ -50,1205 +73,1450 @@ pub trait GpioExt {
...
+///
+/// See [examples/gpio_erased.rs] as an example.
+///
+/// [examples/gpio_erased.rs]: https://github.com/stm32-rs/stm32f3xx-hal/blob/v0.6.0/examples/gpio_erased.rs
+pub type PXx<MODE> = Pin<Gpiox, Ux, MODE>;
+
+macro_rules! modify_at {
+    ($xr:expr, $bits:expr, $i:expr, $val:expr) => {
+        $xr.modify(|r, w| {
+            w.bits(r.bits() & !(u32::MAX >> 32 - $bits << $bits * $i) | $val << $bits * $i)
```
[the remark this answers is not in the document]
Sh3Rm4n: I had to take some time to understand it, but i'd suggest to write the macro like this and comment it a bit more, so that it is easier to parse. Without the knowledge of the operator evaluation order, this is hard to parse.

```suggestion
            let mask = !(u32::MAX >> (32 - $bits) << ($bits * $i));
            let value = $val << $bits * $i;
            w.bits(r.bits() & mask | value)
```

Sh3Rm4n (changes requested):
Wow, this PR is huge. :eyes:  Both in code change, but also implementation wise. 

Props to you, to unclutter the macro and leveraging the rust type-system and static polymorphism techniques, to not rely too heavily on macros to hold it all together. 

Your extensive use of traits not only move most of the dispatch logic out of the macro, but actually makes it easier to read through the gpio implementation itself. 

The interrupt implementation is a very welcomed additions as well. 

Props to you  and many thanks for this implementation. :pray: 

I've added mostly nits as review comments, but design wise, this looks perfect. I'm not able to find any critiques, as I couldn't have done it any better than you. As if I could even do this myself :laughing:

src/gpio.rs
```diff
@@ -50,1205 +73,1450 @@ pub trait GpioExt {
...
 
-/// Alternate function 0 (type state)
-pub struct AF0;
+/// Marker trait for readable pin modes
+pub trait Readable {}
 
-/// Alternate function 1 (type state)
-pub struct AF1;
+/// Marker trait for slew rate configurable pin modes
+pub trait OutputSpeed {}
```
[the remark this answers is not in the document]
Piroro-hs: Users can't implement these marker traits to pin mode typestates thanks to Rust's trait orphan rules.

src/gpio.rs
```diff
@@ -50,1205 +73,1450 @@ pub trait GpioExt {
...
-pub struct Floating;
-/// Pulled down input (type state)
-pub struct PullDown;
-/// Pulled up input (type state)
-pub struct PullUp;
-
-/// Output mode (type state)
-pub struct Output<MODE> {
-    _mode: PhantomData<MODE>,
+use private::{Afr, GpioRegExt, Moder, Ospeedr, Otyper, Pupdr};
```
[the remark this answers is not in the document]
Piroro-hs: To use functions defined in these traits, we need this use statement :(

src/gpio.rs
```diff
@@ -50,1205 +73,1450 @@ pub trait GpioExt {
...
+                use super::{
+                    Input, Output, Analog, PushPull, OpenDrain,
+                    IntoAf0, IntoAf1, IntoAf2, IntoAf3, IntoAf4, IntoAf5, IntoAf6, IntoAf7,
+                    IntoAf8, IntoAf9, IntoAf10, IntoAf11, IntoAf12, IntoAf13, IntoAf14, IntoAf15,
+                    AF0, AF1, AF2, AF3, AF4, AF5, AF6, AF7, AF8, AF9, AF10, AF11, AF12, AF13, AF14, AF15,
+                };
+
+                #[allow(unused_imports)]
+                use typenum::{
+                    U0, U1, U2, U3, U4, U5, U6, U7, U8, U9, U10, U11, U12, U13, U14, U15
```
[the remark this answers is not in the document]
Piroro-hs: Used to implement partially erased pin type aliases.
```rust
pub type $PXi<MODE> = Pin<$Gpiox, $Ui, MODE>;
```

src/serial.rs
```diff
@@ -46,42 +48,44 @@ pub unsafe trait TxPin<USART> {}
 /// RX pin - DO NOT IMPLEMENT THIS TRAIT
 pub unsafe trait RxPin<USART> {}
 
-unsafe impl TxPin<USART1> for gpioa::PA9<AF7> {}
-unsafe impl TxPin<USART1> for gpiob::PB6<AF7> {}
-unsafe impl TxPin<USART1> for gpioc::PC4<AF7> {}
-unsafe impl RxPin<USART1> for gpioa::PA10<AF7> {}
-unsafe impl RxPin<USART1> for gpiob::PB7<AF7> {}
-unsafe impl RxPin<USART1> for gpioc::PC5<AF7> {}
+unsafe impl<OTYPE> TxPin<USART1> for gpioa::PA9<AF7<OTYPE>> {}
```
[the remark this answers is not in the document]
Piroro-hs: Allowing open-drain configuration makes it possible to use the interface on high voltage bus'es using a simple pull-up, so we should keep it. It is also required for USART half-duplex mode (currently not supported in hal, though).

src/gpio.rs
```diff
@@ -50,1205 +73,1450 @@ pub trait GpioExt {
...
+            w.bits(r.bits() & !(u32::MAX >> 32 - $bits << $bits * $i) | $val << $bits * $i)
+        });
+    };
+}
 
+impl<GPIO, INDEX, MODE> Pin<GPIO, INDEX, MODE>
+where
+    INDEX: Unsigned,
+{
+    /// Erases the pin number from the type
```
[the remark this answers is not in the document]
Piroro-hs: Is it ok to use single letter `impl<G, I, M>`? `Gpio` and `Index` will collide with trait names 😥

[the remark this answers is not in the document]
Sh3Rm4n: Hm. That is a good point. I don't really know the best solution. Maybe a namespace aka modules would help? I would prefer changing the name(space) of the traits though. So we can distinguish trait `trait::Gpio` vs. type `type:Gpio`? 

Do you have better naming ideas?

src/serial.rs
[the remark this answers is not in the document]
Sh3Rm4n: Good point. And good thinking in the future. Let's keep it as is.

src/gpio.rs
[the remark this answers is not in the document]
Sh3Rm4n: Updated to stm32f3 v0.13 in  #211

examples/gpio_interrupts.rs
```diff
@@ -0,0 +1,86 @@
...
+    let mut user_button = gpioa
+        .pa0
+        .into_pull_down_input(&mut gpioa.moder, &mut gpioa.pupdr);
+    user_button.make_interrupt_source(&mut syscfg);
+    user_button.trigger_on_edge(&mut exti, Edge::Rising);
+    user_button.enable_interrupt(&mut exti);
+    // Moving ownership to the global BUTTON so we can clear the interrupt pending bit.
+    cortex_m::interrupt::free(|cs| *BUTTON.borrow(cs).borrow_mut() = Some(user_button));
+
+    unsafe { NVIC::unmask(Interrupt::EXTI0) };
```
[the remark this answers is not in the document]
Piroro-hs: `nvic(&self) -> pac::Interrupt` method is added.

src/gpio.rs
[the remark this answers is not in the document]
Piroro-hs: Naming collision resolved by moving marker traits under `marker` module :)

Sh3Rm4n (approved):
Minor tiny nit. 

Overall this looks fantastic. Let me know if you want me to press the merge button :grinning:

Piroro-hs:
Updated (add comments and move `IntoAf..` traits to `marker` module).
This is ready to merge :)

Sh3Rm4n:
Nice. Let's merge it :steam_locomotive: 

Thanks for your hard work, again! :)

Issue #191, opened 2021-01-10
Use embedded-time instead of custom implementation


wbuck:
When writing a driver for some device (say a motor driver) and attempt to program against the `trait`'s in the [embedded-hal](https://github.com/rust-embedded/embedded-hal) you can run into issues when using a `peripheral` that requires types defined in [time.rs](https://github.com/stm32-rs/stm32f3xx-hal/blob/master/src/time.rs).

Because of this the driver now has a dependency on the `stm32f3xx-hal` device crate.
This of course is unwanted as it would be nice to create a `generic` device driver that doesn't depend on any specific device crate.

Example:

    pub struct Driver<T>
    where 
        T: CountDown
    { 
        timer: T,
    }

    impl<T> Driver<T> 
    where 
        T: CountDown
    {
        pub fn new( timer: T ) -> Self { Driver { timer }  }

        pub fn start( &self ) { 
             self.timer.start( /* This requires Hertz which is defined in time.rs of the stm32f3xx-hal crate */ );
        }
    }

Instead of using the units/functionality defined in [time.rs](https://github.com/stm32-rs/stm32f3xx-hal/blob/master/src/time.rs) could the units/functionality defined in [embedded-time](https://github.com/FluenTech/embedded-time) be used instead? This would allow for a truly generic driver (assuming other device crates got on board).

This is probably more of community issue as a lot of device crates use their own custom `time.rs` implementation.

Below are two issues regarding this for the [embedded-hal](https://github.com/rust-embedded/embedded-hal) crate.
It sounded like the consensus was to have device crates get on board with using some centralized source for using time based units and functionality.

[embedded-hal #221](https://github.com/rust-embedded/embedded-hal/issues/211)
[embedded-hal #59](https://github.com/rust-embedded/embedded-hal/issues/59)

wbuck:
Just noticed that there was already a request for this:

[#191](https://github.com/stm32-rs/stm32f3xx-hal/issues/191)

Sh3Rm4n:
Closed by https://github.com/stm32-rs/stm32f3xx-hal/pull/192

Issue #185, opened 2020-12-18
Leverage embedded-time


Sh3Rm4n:
We have some extension traits, which define times and frequencies in [`src/time.rs`](https://github.com/stm32-rs/stm32f3xx-hal/blob/master/src/time.rs). 

This module is mostly a 1:1 copy, which can be found in almost every `stm32f*-hal`.
It mostly used internally, but gets also exposed, to configured clocks and timers. 

Wouldn't it be nicer to use [`embedded-time`](https://github.com/FluenTech/embedded-time) to replace the current [`time`](https://github.com/stm32-rs/stm32f3xx-hal/blob/master/src/time.rs) module? 

Maybe we can leverage some more features of this crate.

wbuck:
So I forked the project, removed the custom `time` module and replaced it with [embedded-time](https://github.com/FluenTech/embedded-time/). It was almost a drop-in replacement. I altered the `serial` constructor to take a `Baud` type instead of the original `Bps`.

I only built it for the `stm32f303xc` but it did build with no issues.
I'm fairly new to `Rust` so I'm unsure what the expectation is regarding re-exporting a third party crate.
So to minimize changes I added the following to `lib.rs`

    pub use embedded_time as time;

David-OConnor:
Please make a PR.

strom-und-spiele:
@Sh3Rm4n thanks for proposing this. And thanks to @wbuck for giving a nice example on how to "leverage some more features of this crate" in  #191. 
Reading about its fixed-point arithmetic under the hood made me like this purpose even more :+1:

Sh3Rm4n:
Closed by https://github.com/stm32-rs/stm32f3xx-hal/pull/192

PR #166, opened 2020-11-04
Added DAC, low-power, alternate clocks, interrupts module. Added seri…

David-OConnor:
- DAC
- Low-power
- Clocks (Set prescalers instead of speed approach)
- External interrupts
- RTC wakeup, and bugfixes
- Serial example
- Feature additions to timer
- Fix for I2C repeating start

I will no longer manage more than a few branches at a time.

This branch will be published periodically as [stm32f3xx-hal-v2](https://crates.io/crates/stm32f3xx-hal-v2).

PR #192, opened 2021-01-11
Feature/embedded time

wbuck:
Replaces the custom [time](https://github.com/stm32-rs/stm32f3xx-hal/blob/master/src/time.rs) module with [embedded-time](https://github.com/FluenTech/embedded-time).

This change addresses both [#191](https://github.com/stm32-rs/stm32f3xx-hal/issues/191) and [#185](https://github.com/stm32-rs/stm32f3xx-hal/issues/185).

This change will allow folks to create custom device drivers without depending on a specific device crate for their custom duration/rate types.


David-OConnor:
A step further is getting rid of `.hz` internally.

wbuck:
@David-OConnor I was thinking the exact same thing.
I was wondering if using the `Generic<u32>` rate would be the way to go.

In that case interfaces would have to be changed (and not only the arguments).
The return types would need to be changed to return a `Result` in the event someone passed in a value that could not be safety converted into `Hertz`.

Unless panicking is acceptable in that case.

strom-und-spiele:
I like the `Result` idea; it'd be nice if we could get rid of the (rather ugly) lines like `Hertz::try_from(3u32.MHz()).unwrap()`  and still keep a nice way to catch conversion errors.

wbuck:
@strom-und-spiele agreed, I didn't like have `Hertz::try_from(3u32.MHz()).unwrap()` felt ugly and non-intuitive. 

Instead, something like this maybe?

    pub fn sysclk( mut self, freq: Generic<u32> ) -> Result<Self, ConversionError> {
        self.sysclk = Some( Hertz::<u32>::try_from( f )?.0 );
        Ok( self )
    }

This looks a little bit better.

    let clocks = rcc
        .cfgr
        .use_hse( 8u32.MHz( ).into( )? )
        .sysclk( 48u32.MHz( ).into( )? )
        .pclk1( 24u32.MHz( ).into( )? )
        .freeze( &mut flash.acr );

strom-und-spiele:
Nice. Thanks for thinking this over and proposing the `Rate` and `TryInto` traits. Looks pretty :sparkles:
 I'll try to give the rest of your PR a read but I guess @Sh3Rm4n should also have a look, as he brought the idea forward (as you mentioned above).

src/i2c.rs
```diff
@@ -451,7 +451,7 @@ macro_rules! i2c {
                 fn clock(clocks: &Clocks) -> Hertz {
                     // NOTE(unsafe) atomic read with no side effects
                     match unsafe { (*RCC::ptr()).cfgr3.read().$i2cXsw().variant() } {
-                        I2C1SW_A::HSI => 8.mhz().into(),
+                        I2C1SW_A::HSI => Hertz(8_000_000),
```
strom-und-spiele, line 454: Whats the motivation to use Hertz instead of Megahertz here?

src/prelude.rs
```diff
@@ -6,7 +6,8 @@ pub use crate::flash::FlashExt as _stm32f3xx_hal_flash_FlashExt;
 pub use crate::gpio::GpioExt as _stm32f3xx_hal_gpio_GpioExt;
 pub use crate::hal::prelude::*;
 pub use crate::rcc::RccExt as _stm32f3xx_hal_rcc_RccExt;
-pub use crate::time::U32Ext as _stm32f3xx_hal_time_U32Ext;
+pub use embedded_time::duration::*;
+pub use embedded_time::rate::*;
```
strom-und-spiele: If I'm still up to date, this way of writing it is preferred:
`pub use embedded_time::{duration::*, rate::*};`

src/time.rs
```diff
@@ -1,5 +1,6 @@
 //! Time units
 
+/*
```
strom-und-spiele: simply delete it instead of commenting it out. If we feel nostalgic, we can always go back and look at it - that's the beauty of git ;)

examples/can.rs
```diff
@@ -31,10 +34,10 @@ fn main() -> ! {
 
     let _clocks = rcc
         .cfgr
-        .use_hse(32.mhz())
-        .sysclk(32.mhz())
-        .pclk1(16.mhz())
-        .pclk2(16.mhz())
+        .use_hse(Hertz::try_from(32u32.MHz()).unwrap())
```
strom-und-spiele, line 37: https://github.com/stm32-rs/stm32f3xx-hal/pull/192#issuecomment-758057396 :+1:

examples/i2c_scanner.rs
```diff
@@ -33,7 +36,14 @@ fn main() -> ! {
         gpiob.pb6.into_af4(&mut gpiob.moder, &mut gpiob.afrl), // SCL
         gpiob.pb7.into_af4(&mut gpiob.moder, &mut gpiob.afrl), // SDA
     );
-    let mut i2c = hal::i2c::I2c::new(dp.I2C1, pins, 100.khz(), clocks, &mut rcc.apb1);
+
+    let mut i2c = hal::i2c::I2c::new(
+        dp.I2C1,
+        pins,
+        Hertz::try_from(100u32.kHz()).unwrap(),
```
strom-und-spiele, line 43: https://github.com/stm32-rs/stm32f3xx-hal/pull/192#issuecomment-758057396 :+1:

examples/pwm.rs
```diff
@@ -28,7 +30,10 @@ fn main() -> ! {
     // Configure our clocks
     let mut flash = dp.FLASH.constrain();
     let mut rcc = dp.RCC.constrain();
-    let clocks = rcc.cfgr.sysclk(16.mhz()).freeze(&mut flash.acr);
+    let clocks = rcc
+        .cfgr
+        .sysclk(Hertz::try_from(16u32.MHz()).unwrap())
```
strom-und-spiele, line 35: https://github.com/stm32-rs/stm32f3xx-hal/pull/192#issuecomment-758057396 :+1:

examples/spi.rs
```diff
@@ -24,9 +27,9 @@ fn main() -> ! {
 
     let clocks = rcc
         .cfgr
-        .use_hse(8.mhz())
-        .sysclk(48.mhz())
-        .pclk1(24.mhz())
+        .use_hse(Hertz::try_from(8u32.MHz()).unwrap())
```
strom-und-spiele, line 30: https://github.com/stm32-rs/stm32f3xx-hal/pull/192#issuecomment-758057396 :+1:

examples/spi.rs
```diff
@@ -43,7 +46,7 @@ fn main() -> ! {
         dp.SPI1,
         (sck, miso, mosi),
         spi_mode,
-        3.mhz(),
+        Hertz::try_from(3u32.MHz()).unwrap(),
```
strom-und-spiele, line 49: https://github.com/stm32-rs/stm32f3xx-hal/pull/192#issuecomment-758057396 :+1:

examples/usb_serial.rs
```diff
@@ -26,10 +29,10 @@ fn main() -> ! {
 
     let clocks = rcc
         .cfgr
-        .use_hse(8.mhz())
-        .sysclk(48.mhz())
-        .pclk1(24.mhz())
-        .pclk2(24.mhz())
+        .use_hse(Hertz::try_from(8u32.MHz()).unwrap())
```
strom-und-spiele, line 32: https://github.com/stm32-rs/stm32f3xx-hal/pull/192#issuecomment-758057396 :+1:

tests/rcc.rs
```diff
@@ -37,14 +38,14 @@ mod tests {
 
         let clock = rcc
             .cfgr
-            .use_hse(8.mhz())
-            .sysclk(15.mhz())
+            .use_hse(Hertz::try_from(8u32.MHz()).unwrap())
```
strom-und-spiele, line 41: https://github.com/stm32-rs/stm32f3xx-hal/pull/192#issuecomment-758057396 :+1:

tests/rcc.rs
```diff
@@ -64,12 +65,12 @@ mod tests {
                     feature = "stm32f303xe",
                     feature = "stm32f398",
             ))] {
-                let clock = rcc.cfgr.sysclk(72.mhz()).freeze(&mut flash.acr);
+                let clock = rcc.cfgr.sysclk(Hertz::try_from(72u32.MHz()).unwrap()).freeze(&mut flash.acr);
```
strom-und-spiele, line 68: https://github.com/stm32-rs/stm32f3xx-hal/pull/192#issuecomment-758057396 :+1:

tests/rcc.rs
```diff
@@ -78,12 +79,12 @@ mod tests {
                 // and the resolution is therefor lower.
                 // Because of the implementation the clock is then approximated to
                 // the highest possible value (64 Mhz).
-                let clock = rcc.cfgr.sysclk(67.mhz()).freeze(&mut flash.acr);
+                let clock = rcc.cfgr.sysclk(Hertz::try_from(67u32.MHz()).unwrap()).freeze(&mut flash.acr);
```
strom-und-spiele, line 82: https://github.com/stm32-rs/stm32f3xx-hal/pull/192#issuecomment-758057396 :+1:

tests/rcc.rs
```diff
@@ -100,14 +101,14 @@ mod tests {
 
         let clock = rcc
             .cfgr
-            .use_hse(8.mhz())
-            .sysclk(32.mhz())
+            .use_hse(Hertz::try_from(8u32.MHz()).unwrap())
```
strom-und-spiele, line 104: https://github.com/stm32-rs/stm32f3xx-hal/pull/192#issuecomment-758057396 :+1:

tests/rcc.rs
```diff
@@ -142,21 +143,21 @@ mod tests {
 
         let clock = rcc
             .cfgr
-            .use_hse(8.mhz())
-            .hclk(48.mhz())
-            .sysclk(48.mhz())
-            .pclk1(12.mhz())
-            .pclk2(12.mhz())
+            .use_hse(Hertz::try_from(8u32.MHz()).unwrap())
```
strom-und-spiele, line 146: https://github.com/stm32-rs/stm32f3xx-hal/pull/192#issuecomment-758057396 :+1:

tests/rcc.rs
```diff
@@ -166,15 +167,15 @@ mod tests {
 
         let clock = rcc
             .cfgr
-            .use_hse(8.mhz())
-            .pclk1(16.mhz())
-            .sysclk(64.mhz())
+            .use_hse(Hertz::try_from(8u32.MHz()).unwrap())
```
strom-und-spiele, line 170: https://github.com/stm32-rs/stm32f3xx-hal/pull/192#issuecomment-758057396 :+1:

tests/rcc.rs
```diff
@@ -189,14 +190,14 @@ mod tests {
 
         let clock = rcc
             .cfgr
-            .use_hse(8.mhz())
-            .sysclk(72.mhz())
+            .use_hse(Hertz::try_from(8u32.MHz()).unwrap())
```
strom-und-spiele, line 193: https://github.com/stm32-rs/stm32f3xx-hal/pull/192#issuecomment-758057396 :+1:

strom-und-spiele (changes requested):
I'm (positively) surprised how little change is needed to adopt embedded-time. IMHO this speaks a lot for adopting it. 
My change requests are minor nits and the conversion feature you proposed in https://github.com/stm32-rs/stm32f3xx-hal/pull/192#issuecomment-758057396.
I personally prefer having a lot of TODO boxes to check (and get a dopamine dose every time I do so) over missing on a change, so if I added redundant comments, the intention was to be helpful not to annoy. :)

src/i2c.rs
wbuck: In this case it is/was the return `type` of that specific function. I didn't want to change it to `MHz` so that I didn't break anything unintentionally.

src/prelude.rs
wbuck: I believe I removed these in the latest commit.

wbuck:
So I have started changing functions taking a frequency to what I proposed above.

    fn func_taking_rate<R>( mut self, freq: R ) -> Result<Self, <R as TryInto<Hertz<u32>>>::Error> 
    where
        R: Rate + TryInto<Hertz<u32>>

I'm lazy so I personally prefer calling this:

    some_func( 8u32.MHz( ) ).unwrap( );

Over this:

    some_func( Hertz::try_from( 8u32.MHz( ).unwrap( ) )

Or this:

    some_func( 8u32.MHz( ).into( ) ).unwrap( );

src/i2c.rs
strom-und-spiele: …and using Megahertz and casting it back doesn't really make it more readable.

strom-und-spiele (approved):

David-OConnor:
Is this finalized? Do you plan to make additional changes?

wbuck:
@David-OConnor There is one last thing I was toying with and it has to do with the `CountDown`/`Periodic` timer. I wanted to see if I could find a more intuitive way to set the `count` value of the timer by using a `Duration` (milliseconds, microseconds, etc) instead of a `Rate` (hertz). 

I'm really unsure about this though and maybe i'll just abandon that idea all together (in that case this PR is finalized).
I wrote a quick example of what I was thinking but I wanted some feedback on it.

This example was written in another toy application so I didn't have access to the `crate private` methods that are defined in the `stm32f3xx-hal` so I just hacked it for testing purposes.

_This is only an example._

**Timer / CountDown implementation**

    pub struct Timer<D, TIM> 
    where
         D: Duration + FixedPoint<T = u32>,
    {
        clocks: Clocks,
        tim: TIM,
        _pantom: PhantomData<D>,
    }

    impl<D> CountDown for Timer<D, TIM1>  
    where
        D: Duration + FixedPoint<T = u32>,
    {
         type Time = D;

        fn start<T>( &mut self, count: T )
        where
            T: Into<Self::Time> 
        {
            self.stop( );

            let frequency: Hertz = ( count.into( ) as D ).to_rate( ).unwrap( );
            let frequency = frequency.0;
            
            // Same as current impl.
        }

        fn wait( &mut self ) -> nb::Result<( ), void::Void> {
           // Same as current impl.
        }
    }

    // Timer impl same as current.

**Instantiating the timer**

    let mut timer = Timer::tim1( p.TIM1, 10u32.microseconds( ), clocks, &mut rcc.apb2 );

**Using the timer**

    timer.start( 50u32.microseconds( ) );

Maybe I've missed the point regarding the `CountDown` timer but when I'm setting a timer I usually like using a duration instead of a rate.

David-OConnor:
Agree on duration instead of rate. The Hz abstraction doesn't even let you set a duration > 1s. 

For an example of how to set duration, ref my PR here: https://github.com/stm32-rs/stm32f3xx-hal/pull/166/files

It doesn't use `CountDown`, but I think that would be nice. see `set_period` for an example on how to setup a time. Check the comments for caveats. I'm n clear on how the existing method works, but it doesn't produce accurate results for me.

```rust

                /// Set the timer period, in seconds. Overrides the period or frequency set
                /// in the constructor.
                /// This allows you to set periods greater than 1hz.
                pub fn set_period(&mut self, period: f32, clocks: &clocks::Clocks) -> Result<(), ValueError> {
                    // PSC and ARR range: 0 to 65535
                    // (PSC+1)*(ARR+1) = TIMclk/Updatefrequency = TIMclk * period
                    // APB1 (pclk1) is used by Tim2, 3, 4, 6, 7.
                    // APB2 (pclk2) is used by Tim8, 15-20 etc.
                    // todo: It appears there's a (fixed?) 2x multiplier on APB1
                    // timers; it's twice `pclk1`. See clocks diagram in RM, or `Clock Configuration`
                    // tool in STM32CubeIDE.
                    let tim_clk = clocks.calc_speeds().pclk1 * 1_000_000. * 2.;

                    // We need to factor the right-hand-side of the above equation (`rhs` variable)
                    // into integers. There are likely clever algorithms available to do this.
                    // Some examples: https://cp-algorithms.com/algebra/factorization.html
                    // We've chosen something quick to write, and with sloppy precision;
                    // should be good enough for most cases.

                    // - If you work with pure floats, there are an infinite number of solutions: Ie for any value of PSC, you can find an ARR to solve the equation.
                    // - The actual values are integers that must be between 0 and 65_536
                    // - Different combinations will result in different amounts of rounding errors. Ideally, we pick the one with the lowest rounding error.
                    // - The aboveapproach sets PSC and ARR always equal to each other.
                    // This results in concise code, is computationally easy, and doesn't limit
                    // the maximum period. There will usually be solutions that have a smaller rounding error.

                    let max_val = 65_535;
                    let rhs = tim_clk * period;

                    // todo: Round instead of cast?
                    let arr = (rhs.sqrt() - 1.) as u16;
                    let psc = arr;

                    if arr > max_val || psc > max_val {
                        return Err(ValueError {})
                    }


                    self.tim.arr.write(|w| unsafe { w.bits(u32::from(arr)) });
                    self.tim.psc.write(|w| unsafe { w.bits(u32::from(psc)) });

                    Ok(())
                }

                /// Reset the countdown; set the counter to 0.
                pub fn reset_countdown(&mut self) {
                    self.tim.cnt.write(|w| unsafe { w.bits(0) });
                }
            }
```

wbuck:
@David-OConnor Yeah I was looking at the issue you brought up regarding timer durations greater than 1 second.

Another idea I've been playing with but isn't even close to prime time is creating a `TimerBuilder`.
The pattern would allow for both the creation of a simple timer (hiding all complex stuff) and a complicated timer.

The plan is to allow the user to configure the `CountDown` timer however they want.

Example:

    let mut timer = TimerBuilder::new( TIM1, &mut rcc ).repeating( ).period( | w | { } ).duration( | w | { } ).build( );

This is still a pie in the sky idea, but I have started to try and map out the potential different branches one might take when building a `CountDown` timer.

David-OConnor:
I like that. I like how the `embedded-hal` CountDown etc traits provide a basic abstraction that will have identical syntax anwhere: Across HALs, between timer, RTC and systick etc. It's deliberately limited in how you can configure it - that's why I agree with your idea of separating simple and complex.

David-OConnor:
Yes

wbuck:
@David-OConnor 

I think there is a small issue in the `set_period` method.
`PCLK1` is only doubled in the event that the `APB1` prescaler (`PPRE1`) is set to anything above 3.

    If (APB1 prescaler=1) x1 else x2

Also, switching the type for the `start` method  from a `Rate` type (`Hertz<u32>`) to a `Duration<u32>` type isn't going to solve the issue you're having (periods > 1s).

I still convert the duration into a frequency `Hertz<u32>` within the start methods body.

    let frequency: Hertz = ( timeout.into( ) as D ).to_rate( ).unwrap( );

Anything greater than 1 second will return 0 for the frequency.
So again, the same issue you're currently having would persist.

Really the only solution I can think of for periods > 1s would be to do what you have already done with the `set_period` method.

David-OConnor:
Good catch on PCLK1. I can confirm that's how it works from the Cube IDE. In my code, it's fixed at 2. You've IDed the correct logic. Also, my code doesn't differentiate between apb1 or apb2, although there's a comment indicating it.

I think the sqrt approach I made works for long durations, but may not be precise enough for short ones, where you'd need to have one of the values (PSC, ARR) smaller than the other.

I can't comment on the >1s issue, since I haven't looked at the Duration type you're using.

David-OConnor:
edit: I understand your >1 s point. It's not about the `Duraction` struct; it's about the PSC and ARR calculation.

Btw: Not in the code I posted, but I think it makes sense to split the PSC and ARR calculation into a standalone, pure function, then call it from `CoutndownDown` and `set_period`.

David-OConnor:
I think the timer clock computation makes more sense in the rcc or clocks module, so you can just do this:
```rust
let tim_clk = clocks.calc_speeds().timclock1 * 1_000_000.;
```

David-OConnor:
I can confirm that the sqrt approach I took is imprecise at high frequencies.

wbuck:
@David-OConnor So I think I'm going to keep the `CountDown` timer the way it currently is (keep this PR fairly simple). 

I would prefer to open a different PR for changing the timer. 

I like your approach of allowing folks to adjust the period but what about the `start` method?.
Currently if someone calls `set_period` and then calls `start` the previously calculated `ARR` and `PSC` values would be blown away.

Someone would have to first call the `ctor` (which in turn calls `start` enabling the timer) then they would have to call `set_period`.

But I guess given how general the `CountDown` trait is, there really is no other simple and clean solution.

David-OConnor:
I think this goes back to your idea of simple and complex: Unfortunately, I don't think there's a way around that limitation with the `CountDown` impl. One way would be raw pointers. Another is to PR `embedded-hal` to add a `change_period` method, but they like to keep those traits as generic as possible.

Sh3Rm4n:
I have not gone through all comments and changes in detail, just wanted to ask if this PR is still kinda WIP or if it is worth to fully review it to get it into the master branch :)

wbuck:
@Sh3Rm4n All work is complete and it’s ready for a full review.

src/i2c.rs
```diff
@@ -111,14 +111,20 @@ macro_rules! busy_wait {
...
+    ) -> Result<Self, <F as TryInto<Hertz<u32>>>::Error>
     where
         I2C: Instance,
         SCL: SclPin<I2C>,
         SDA: SdaPin<I2C>,
-        F: Into<Hertz>,
+        F: Rate + TryInto<Hertz<u32>>,
     {
-        let freq = freq.into().0;
+        let freq = (freq.try_into()? as Hertz).0;
```
Sh3Rm4n, line 127: This is a breaking change and should be noted in the CHANGELOG.md

But I'm not really a fan of this approch, as `F`, which now has to implement `TryInto` is validated inside the function, which in turn just returns the conversion error of `TryInto`. 

I'd suggest, so that the user has to validate the conversion themselves, before passing it into the function:

```suggestion
    pub fn new<F>(
        i2c: I2C,
        pins: (SCL, SDA),
        freq: Hertz,
        clocks: Clocks,
        apb1: &mut APB1,
    ) -> Self
    where
        I2C: Instance,
        SCL: SclPin<I2C>,
        SDA: SdaPin<I2C>,
    {
        let freq = freq.0;
```

src/rcc.rs
```diff
@@ -333,12 +335,12 @@ impl CFGR {
     ///
     /// Will result in a hang if an external oscillator is not connected or it fails to start,
     /// unless [css](CFGR::enable_css) is enabled.
-    pub fn use_hse<F>(mut self, freq: F) -> Self
+    pub fn use_hse<F>(mut self, freq: F) -> Result<Self, <F as TryInto<Hertz<u32>>>::Error>
```
Sh3Rm4n, line 338: Same here, with `Result` breaking the API for a conversion error, which could be done by the user themselves, before passing it into the function. 

I'd like to reserve the change to using `Result` types to actual logic validation checks at runtime, which **can't** possibly be done manually beforehand.

Sh3Rm4n (changes requested):
Wow, this is prettry cleany applicable without to much change. 

I'm only unhappy about the `Result` change. 

Regarding https://github.com/stm32-rs/stm32f3xx-hal/pull/192#issuecomment-758057396, I know that the current clocks generation results into ugly code, but this is on my todo list for a long time. 

```rust
let clocks = rcc
    .cfgr
    .use_hse( 8u32.MHz( ).into( ) )?
    .sysclk( 48u32.MHz( ).into( ) )?
    .pclk1( 24u32.MHz( ).into( ) )?
    .freeze( &mut flash.acr );
```

which is your 2nd proposal, which does look pretty clean to me. It is not as clever as the `TryInto` constrain, but it also does not need to unnecessarily introduce a change in the function signature (aka introducing `Result` as the return type). 

Thanks for your contribution! I think my suggestion is an easy fix. Otherwise this looks pretty good :sparkles:

src/rcc.rs
wbuck: OK, so you'd prefer the functions taking a generic `F` type which are constrained to `Rate + TryInto<u32>` instead be changed to taking a `Hertz` concrete type instead?

In other words, put the onus of the conversion on the user.

So, for example:

This signature:

    pub fn pclk2<F>(mut self, freq: F) -> Result<Self, <F as TryInto<Hertz<u32>>>::Error>
    where
        F: Rate + TryInto<Hertz<u32>>;

Becomes this:

    pub fn pclk2( mut self, freq: Hertz ) -> Self;

Is this what you had in mind?

Sh3Rm4n: Yes, this is what I meant. 

Maybe I'm overlooking something, but I can find no strong argument to introduce a conversion, which can fail inside the function, which error has to be propagated out of it, if the alternative is to just check the failing conversion before passing the argument into the function. 

I know, this has slight ergonomically disadvantages, but I don't think it is too bad. 
Anyways, there is no case, where the conversion into `Hertz` can beforehand, am I correct?

wbuck: Yeah that's totally fine. I had only done that initially to take the burden of the conversion off the user.
And yes, you're correct.

There is another approach as well; I figure I'll throw it out there.

We know that the maximum MCU frequency for the F3 is `74MHz` and we also know that that will comfortably fit in a `u32`. What if instead of returning a `Result` we `panic`? In the case where someone passes in a value that cannot fit into a `u32` they obviously did something wrong as no F3 as far as I know has a clock speed that is capable of `> 4.29GHz`.  

I agree that returning a `Result` in this case doesn't really make sense as what is the user going to do? Start setting different clock frequencies in a loop until finally they get the `Ok`? Probably not I would think.

I'm good either way. I can make the changes tonight after my daughter goes to bed.

Sh3Rm4n: Your approach is mentioning a weakness of the current API I wanted to tackle for ages, but did not had the time now. 
But it's different to a conversion error. The maximum frequency allowed depends heavily on the peripheral, which gets configured. So it is really context dependent. This problem is about value constrains. 
E.g. I started to tackle to problem for the RCC module in https://github.com/stm32-rs/stm32f3xx-hal/pull/153. Though I'm not really happy with the approch and also had no time to further work on it. 

To not change the API for know, `panic`ing is fine, though it should already panic in most (if not all cases), when the frequency is too high. 

The problem regarding this PR is the conversion error though. That is a different error, which will be thrown by `TryFrom` / `TryInto`. This conversion error can be checked by the user itself and should not justify a change on the API surface.

wbuck: OK no problem. What I'll do is I'll roll back what I had done previously with the `TryFrom`/`TryInto` and instead change the the parameter types to take `Hertz`.

src/spi.rs
```diff
@@ -424,9 +424,9 @@ macro_rules! hal {
                     freq: F,
                     clocks: Clocks,
                     apb2: &mut $APBX,
-                ) -> Self
+                ) -> Result<Self, <F as TryInto<Hertz<u32>>>::Error>
```
Sh3Rm4n, line 427: Use missed reverting this part :)

src/spi.rs
```diff
@@ -494,13 +495,18 @@ macro_rules! hal {
 
                 /// Change the baud rate of the SPI
                 pub fn reclock<F>(&mut self, freq: F, clocks: Clocks)
-                    where F: Into<Hertz>
+                -> Result<(), <F as TryInto<Hertz<u32>>>::Error>
```
Sh3Rm4n, line 498: Here as well.

examples/can.rs
```diff
@@ -31,10 +32,10 @@ fn main() -> ! {
 
     let _clocks = rcc
         .cfgr
-        .use_hse(32.mhz())
-        .sysclk(32.mhz())
-        .pclk1(16.mhz())
-        .pclk2(16.mhz())
+        .use_hse(Hertz(32_000_000))
```
Sh3Rm4n, line 35: Hm. Now I know why you were going for the result approch before, I've played around a bit to look for alternatives:

```suggestion
        .use_hse(32_000_000_u32.Hz())
```

```suggestion
        .use_hse(32u32.MHz().try_into().unwrap())
```

or even change the corresponding function input type to 

```rust
pub fn use_hse(mut self, freq: Megahertz) -> Self
```

so you can write 

```suggestion
        .use_hse(32u32.MHz())
```

Definitely for `src.rcc.rs` I think the last suggestions should be used. We know that 72.Mhz() is the limit anyways, and this can be safely converted to Hz in u32 bit range.

Though this constrains the resolution to `1.MHz()`, which I'm fine with in the RCC case, because finer resolutions are not really guaranteed anyways. (Maybe I'm wrong though and do overlook something)

For the rest of the modules, I think it is not as easily applicable. I think we should stick to Hertz in these cases. 
In these  cases I slightly prefer my first suggestion in favor of your current approach. But this is not a strong opinion and your current approach is definitely fine!

src/i2c.rs
```diff
@@ -111,14 +111,20 @@ macro_rules! busy_wait {
...
 impl<I2C, SCL, SDA> I2c<I2C, (SCL, SDA)> {
     /// Configures the I2C peripheral to work in master mode
-    pub fn new<F>(i2c: I2C, pins: (SCL, SDA), freq: F, clocks: Clocks, apb1: &mut APB1) -> Self
+    pub fn new<F>(
+        i2c: I2C,
+        pins: (SCL, SDA),
+        freq: F,
+        clocks: Clocks,
+        apb1: &mut APB1,
+    ) -> Result<Self, <F as TryInto<Hertz<u32>>>::Error>
```
Sh3Rm4n, line 120: This should be reverted to `new(...) -> Self, as suggested before.

Sh3Rm4n (changes requested):
Mostly ergonomic suggestions and pointing out some missed `Result<Self,...>` to `Self` reversions. 

This looks promising. :+1:

src/spi.rs
wbuck: Oh, not sure how I missed these..

examples/can.rs
wbuck: I'll take a look at the `RM` later tonight, just to make sure that won't adversely affect anyone.

wbuck: What are your thoughts on the following:

    /// # Panics
    ///
    /// Panics if conversion from `Megahertz` to `Hertz` produces a 
    /// value greater than `u32::MAX`.
    pub fn use_hse(mut self, freq: Megahertz) -> Self {
        let freq: Hertz = freq.try_into( ).unwrap( );
        self.hse = Some(freq.0);
        self
    }

Doing it like above means minimal internal changes in `rcc`. 
The only change would be to the public API.

CHANGELOG.md
```diff
@@ -7,6 +7,31 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [v0.7.0] - 2021-03-10
+
+### Added
+
+- Replace custom time based units with types defined in the [embedded-time][] crate ([#191])
```
Sh3Rm4n, line 14:
```suggestion
- Replace custom time based units with types defined in the [embedded-time] crate ([#192])
```

examples/i2c_scanner.rs
```diff
@@ -33,7 +34,14 @@ fn main() -> ! {
         gpiob.pb6.into_af4(&mut gpiob.moder, &mut gpiob.afrl), // SCL
         gpiob.pb7.into_af4(&mut gpiob.moder, &mut gpiob.afrl), // SDA
     );
-    let mut i2c = hal::i2c::I2c::new(dp.I2C1, pins, 100.khz(), clocks, &mut rcc.apb1);
+
+    let mut i2c = hal::i2c::I2c::new(
+        dp.I2C1,
+        pins,
+        100_000u32.kHz().try_into().unwrap(),
```
Sh3Rm4n, line 41: Shouldn't this be:

```suggestion
        100u32.kHz().try_into().unwrap(),
```

src/i2c.rs
```diff
@@ -133,16 +130,16 @@ impl<I2C, SCL, SDA> I2c<I2C, (SCL, SDA)> {
         // t_SYNC1 + t_SYNC2 > 4 * t_I2CCLK
         // t_SCL ~= t_SYNC1 + t_SYNC2 + t_SCLL + t_SCLH
         let i2cclk = I2C::clock(&clocks).0;
-        let ratio = i2cclk / freq - 4;
-        let (presc, scll, sclh, sdadel, scldel) = if freq >= 100_000 {
+        let ratio = i2cclk / freq.0 - 4;
```
Sh3Rm4n, line 133: Seems like `embedded-time` has an [`integer`](https://docs.rs/embedded-time/0.10.1/embedded_time/rate/trait.Rate.html#get-the-integer-part) method, which could be used instead. 

```suggestion
        let ratio = i2cclk / freq.integer() - 4;
```

Make the code a little better to understand but also more verbose. I'm fine, if you want to keep it as is for now.

Sh3Rm4n (changes requested):
Some minor nits, but this looks really good now overall! :)

examples/i2c_scanner.rs
wbuck: Oh, good catch.

Sh3Rm4n (approved):
Nice. Looks good to me! Thank you for your work and patience.

Issue #195, opened 2021-01-15
Low resolutions of pwm affect the frequency


seanybaggins:
using this code
```rust
use cortex_m_semihosting::{dbg, hprintln};
use panic_semihosting as _;

use stm32f3xx_hal as hal;
use hal::prelude::*;
use hal::pwm;

use cortex_m::asm;

use cortex_m_rt::entry;

#[cfg_attr(not(test), entry)]
fn main() -> ! {
    let device_peripherals = hal::pac::Peripherals::take().unwrap();
    // Setting the clock frequency to the maximum supported value
    let mut flash = device_peripherals.FLASH.constrain();
    let mut reset_and_control_clock = device_peripherals.RCC.constrain();
    let clocks = reset_and_control_clock
        .cfgr
        .sysclk(16.mhz())
        .freeze(&mut flash.acr);

    // Configuring PWM
    let resolution = 250;
    let frequency = 1.hz();
    let pwm_channel_no_pins =
        pwm::tim16(device_peripherals.TIM16, resolution, frequency, &clocks);
    let mut gpiob = device_peripherals.GPIOB.split(&mut reset_and_control_clock.ahb);
    let pb8 = gpiob.pb8.into_af1(&mut gpiob.moder, &mut gpiob.afrh);
    let mut pwm_channel = pwm_channel_no_pins
        .output_to_pb8(pb8);
    pwm_channel.set_duty(pwm_channel.get_max_duty() / 2);
    pwm_channel.enable();

    loop{}
```

I get a pwm signal that I expect. A 50% duty cycle with a 1 second period.

If I change `resolution = 200` then I get a signal that has a period of about 1/10 of the original.

seanybaggins:
Are there any tradeoffs to setting the resolution to u16::MAX. If not perhaps the solution does not need to be exposed to the user of the library.

IamfromSpace:
Hi @seanybaggins, original author of that particular module here.  The issue you're running into is a variant of the one mentioned in #138.  In that one, the resolution is set so high that the prescaler becomes too coarse to yield a reasonably close approximation to the desired frequency.  In your case, the resolution is set so low that prescaler overflows.

[...]

Issue #138, opened 2020-09-09
Tim2: Trick to setting fine resolutions?


David-OConnor:
Hi. I noticed after testing with an oscillosope, I'm having trouble setting fine resolutions using TIM2. Code:

```rust
    let tim2_channels = tim2(
        dp.TIM2,
        160_000,  // resolution of duty cycle
        94.hz(), // frequency of period
        &clocks, // To get the timer's clock speed
    );

    let mut pwm0 = tim2_channels.0.output_to_pa0(pwm0_pin);
    pwm0.set_duty(pwm0.get_max_duty() / 2); // 50% duty cycle
    pwm0.enable();
```
The scope reports a frequency of 100Hz. If set anywhere between 100 and 60, it outputs 100Hz. If set to 50, it outputs 50Hz. Above 100, and I don't see any output. Are there any gotchas here? Thank you.

ra-kete:
Looks like some kind of rounding issue. The relevant code is [here in the pwm module](https://github.com/stm32-rs/stm32f3xx-hal/blob/788aedbb179ecb9bbdcddfdf110dd5896fb38bac/src/pwm.rs#L437).

It's pretty clear that the calculation of the timer values depends on the configured clocks, so may be worth playing around with those. Maybe increasing PCLK1 also increases the accuracy of the output frequency?

I'm afraid I won't be of much help here. The STM timers are extremely complex and I can't claim to understand them.

David-OConnor:
Ooh good point! I have been experimenting with the clocks, although main the `.sysclk()`. I wasn't able to get more precision that way, but haven't tried `PCLK`. Will try that and report back.

I need to dig into the ref man on timers and clocks in general on this. I've never used a scope before or tried configuring them.

David-OConnor:
After more experimenting, it seems several factors go into the frequency output. The clock config as you pointed out (Which I don't understand enough to comment on), and the `resolution of duty cycle` arg, ie `160_000` above, have interplay with the `94.hz()` in ways I don't understand. 

`resolution of duty cycle` appears to be the contents of the `pac::tim2::arr` register.

IamfromSpace:
@David-OConnor Hi David, I originally wrote much of this module, and for the most part this speaks to the challenge in finding a natural interface here.  Though the examples themselves may be "too optimized," in that if you fiddle with them, yeah, they don't do what you expect :(

[...]

Issue #190, opened 2021-01-02
Timers with period > 1s


David-OConnor:
Hi. Is it possible to implement timers with period > 1s? It seems that the `.hz()` api only takes integers.

If not, are you open to a PR?

Sh3Rm4n:
I'm not sure, if it is really possible, but it should. 

Some pointers / ideas:

1. Hardware: Does the timer support frequencies lower than 1 Hz?
2. The [`Into<Hertz>`](https://github.com/stm32-rs/stm32f3xx-hal/blob/c68c36c03e0e33699b3b0c9acc3f8d80f5a25cd4/src/timer.rs#L141) has to be adjusted, as this currently limits the maximal timer period to 1 s.
   - Traits and `From<...>` conversions can be found [here](https://github.com/stm32-rs/stm32f3xx-hal/blob/c68c36c03e0e33699b3b0c9acc3f8d80f5a25cd4/src/time.rs). Maybe extending the [`U32Ext`](https://github.com/stm32-rs/stm32f3xx-hal/blob/master/src/time.rs#L27-L43) implementation is useful? 
   - This is possible not enough, and this constraint has to be reworked to also support values in the `< 1 Hz` / `> 1 s` range 
3. (maybe) The [prescaler and auto-reload value calculation](https://github.com/stm32-rs/stm32f3xx-hal/blob/master/src/timer.rs#L100-L109) has to be adjusted
    - Where `PSC` is the prescaler scaling down the peripheral clock which is connected to the timer block
    - And `ARR` is the auto-reload value, which holds the amount of timer ticks until an event / interrupt fires
    - The maximum value for `ARR` is the maximum possible length for a timer period in Hz, assuming `PSC` scales the peripheral clock down to 1 Hz

David-OConnor:
1. Yep! 

2. I think we'd need to bypass the `Into<Hertz>`, or fork it; I think it's the root cause. I remember looking at the source code a while back; I didn't understand how it worked. Reliance on it is the issue. I'd be happy setting period or frequency with a f32 argument, but that's a breaking change.

3. That's the core of it; PSC and ARR, in conjunction with clock speed are the only thing that determines period. For a given period, you set a combination of PSC and ARR. There is more than one combo for each period; we could pick arbitrarily.  I need to dive into the docs more, but IIIRC the period is (ARR+1) * (PSC+1) * the timer's clock speed.

[...]

PR #196, opened 2021-01-16
(Breaking) Add more res/freq configurations to PWM

IamfromSpace:
This is still a bit in draft form, but I was wanting to get more feedback sooner.  Also, this relates very closely to other timer discussions that I've seen going about recently, but haven't closely followed, so I wanted to makes sure that this aligned with what other folks were seeing for things like this going forward.

Issues of intuitive configuration have come up in both #138 and now #195.  A single path to configuring the registers is probably not sufficient, as it seems there will always be setups that don't quite line up right.  This adds a trait to get the final register values based on the current timer clock.  A power user might just send in these values directly (the trait impl of just the identity function), while there may be a variety of other possible structs that allow the values to be calculated differently.  Ideally, picking a config style that matches the intent more frequently leads to correct results.

### DIFF
--- a/src/pwm.rs
+++ b/src/pwm.rs
@@ -241,6 +241,43 @@ pub struct PwmChannel<X, T> {
     pin_status: PhantomData<T>,
 }
 
+pub struct ExactArrAndPsc<r> {
+    pub arr: r,
+    pub psc: u16,
+}
+
+pub trait ClocksToArrAndPsc {
+    type Arr;
+
+    fn timer_freq_to_arr_and_psc(self, timer_freq: u32) -> ExactArrAndPsc<Self::Arr>;
+}
+
+impl<arr> ClocksToArrAndPsc for ExactArrAndPsc<arr> {
+    type Arr = arr;
+
+    fn timer_freq_to_arr_and_psc(self, timer_freq: u32) -> ExactArrAndPsc<Self::Arr> {
+        self
+    }
+}
+
+pub struct ResAndFreq<r> {
+    resolution: r,
+    frequency: Hertz,
+}
+
+impl<arr: Into<u32>> ClocksToArrAndPsc for ResAndFreq<arr> {
+    type Arr = arr;
+
+    fn timer_freq_to_arr_and_psc(self, timer_freq: u32) -> ExactArrAndPsc<Self::Arr> {
+        // calculate the pre-scaler
+        let prescale_factor = timer_freq / self.resolution.into() / self.frequency.0;
+        ExactArrAndPsc {
+            arr: self.resolution,
+            psc: prescale_factor as u16 - 1,
+        }
+    }
+}
+
 macro_rules! pwm_timer_private {
     ($timx:ident, $TIMx:ty, $res:ty, $apbxenr:ident, $apbxrstr:ident, $pclkz:ident, $timxrst:ident, $timxen:ident, $enable_break_timer:expr, [$($TIMx_CHy:ident),+], [$($x:ident),+]) => {
         /// Create one or more output channels from a TIM Peripheral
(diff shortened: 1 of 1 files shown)


David-OConnor:
Thoughts on returning as part of the result, a percent (or portion) off from requested?

IamfromSpace:
The idea here is that we expose multiple strategies to a user, ideally they gravitate to the one that has the fewest leaks.

So another one to add might be MaximizeResolution where a frequency tolerance is suppied.

Another (I hadn’t thought of this before your comments in #195) could be MaximizeFreqAccuracy, where ARR only gets what’s left over (or nothing).  That works great as a general timer approach.

In theory, this is a good place in the docs where we can both describe typical strategies, their use-cases, and pitfalls.

It may make sense to return a result, so certain strategies can signal when/why the fail, but my expectation is that most consumers (the timers) wouldn’t really be able to do much besides panic at that point anyway.

IamfromSpace:
Just added an interesting approach that seems to check a lot of boxes.  The `PrioritizeResolution` strategy tries to find the largest resolution that doesn't overly sacrifice frequency accuracy.  With such a high resolution as well, it's likely that the duty cycle can be controlled finely enough that individual steps don't accumulate their own error.

Here's an example calculation of the error bounds, from in the comments (should likely be in the docs as that gets ironed out):
```
// Example:
//   - Fi = 72MHz
//   - Fo = 50Hz
//   - Max Resolution = 2^16 (standard timer)
//
// Then:
//   - Worst Case Error ~= 0.00146%
//   - Worst Case Frequency ~= 50.000728Hz
//   - Actual Frequency ~= 0.000833%
//   - Actual Frequency ~= 50.0004167Hz
```
Also, whenever the division doesn't need to be factored (always the case for u32 high resolution timers), the full value goes into ARR and there's no (extra) loss in accuracy from this approach.  That's >1098Hz for standard timers.

Thoughts on this?  Hopefully it seems appealing.

IamfromSpace:
Just added a `PrioritizeFreq` strategy which, @David-OConnor, is similar to your square root approach.  However, this looks for smallest factors first, in hopes that we find an imbalanced factoring, so we can arrive at a large resolution too.

This just uses simple wheel factorization because it has O(1) space complexity, with minimal memory usage, and is deterministic.  The classic approaches here for smaller numbers are a lookup table (way too large) or probabilistic algorithms.  It seemed inappropriate to add randomness here, not knowing the ordering of factors seems like a major downside in getting nice resolutions, and it seems like most inputs in practice will have many small factors (lots of 2s and 5s), so our wheel should work well in many cases.

External package options seemed mostly geared around cryptography, mostly didn't support no_std, or still consumed a lot of memory via allocators.  I tried to do my research here, but I'm not a number theory expert, so there may yet be better ways to do this.  The hardware constraints seem to severely limit options though.

Lastly, we don't get the _optimal_ resolution, because by always taking the smallest factor, they're might be a combination with a larger factor that's better (ex. factors are 2,2,3 where 4 makes the resolution small enough, but so does 3).  To solve that (something like `PrioritizeFreqThenMaximizeRes`) we encounter the [Subset Product Problem which is NP-Complete](https://cgi.csc.liv.ac.uk/~ped/teachadmin/COMP202/annotated_np.html), so I'm not even sure it's worth recommending if we added it.

IamfromSpace:
From a general standpoint, this is strategy is a good one for timers not in PWM mode as well, so it's worth exploring lifting this up a level as a general timer configuration approach.  Notably, this could also answer #190 because a variant of these strategies could accept a ratio.  Something like:

```
OptimizeFreqRatio {
    frequency_numerator: u32,
    frequency_denominator: u32,
}
```

So you could represent a period of 9.5s with something like:
```
OptimizeFreqRatio {
    frequency_numerator: 19,
    frequency_denominator: 2,
}
```

I can take a look at extending the code here to support these as strategies as well.

Does this seem like a good solve?

David-OConnor:
I'll add your wheel factorization approach to my forks here and on L4 after some testing

IamfromSpace:
Great—note this is currently untested.  My immediate goal has been to prove things out and make sure I don’t work in a bubble and miss things that others would have identified up front.  I’ll prioritize testing this out next and let you know when it’s good to go.  In theory, it could come in as a separate PR/module.  The good thing is that these kinds of things unit test quite nicely.

That’s good callout on reuse.  This _technically_ can support it currently but it’s not immediately obvious.  At the end of the day, you need the register values.  These strategies all have a trait which allow you to calculated them.  However, the register values themselves satisfy the trait implicitly.

So my expectation for usage was to pass in strategies (which especially makes sense for PWM, where you are likely to configure once). However, it’s also possible to pre-calculate manually, but simply applying the trait and holding the resulting register values for later—since they themselves satisfy the trait to configure a timer.

A wrinkle there is that the resulting calculation is only guaranteed to be correct per timer.  In the event that you pass in a strategy for a timer that differs in input frequency from the original, you’ll get bad results.  Maybe we could add a phantom type for the timer it was calculated for?  Or maybe there are other alternatives?

Sh3Rm4n:
Thank you for working on this. I haven't looked to deeply into this, so I can't add anything constructive on the design site of things yet. But I really like the aim to provide multiple solutions to the user\1

If a review of the code is needed, just ping me.

PR #198, opened 2021-01-26
[WIP] I2C with interrupts

clebi:
It is a first attempt to implement I2C communications using interrupts. I know that this is not ready for merge but I wanted to have a feedback on my work on I2C and see if I am going into the right direction. Thruthfully, I am not an experienced embedded developer. 
As of now, I don't think there is someting in embedded-hal define to support i2c communication with interrupts.

The interrupt function is responsible for computing the state of the I2C device on every interruption. For the moment, calling this function is the responsability of the consumer.

You can find an example of use here: [example](https://github.com/clebi/i2c_try/blob/master/src/main.rs). The example is using cortex-m-rtic. In this example the interrupt function is called for I2C1_EV_EXTI23 and I2C1_ER.

### DIFF
--- a/src/i2cint.rs
+++ b/src/i2cint.rs
@@ -1,0 +1,377 @@
+//! Inter-Integrated Circuit (I2C) bus interruption
+
+use core::fmt;
+use crate::{i2c::{I2c, Instance, SclPin, SdaPin}};
+use crate::rcc::{Clocks, APB1};
+use crate::time::Hertz;
+
+/// I2c errors.
+#[derive(Debug, Copy, Clone)]
+pub enum I2cError {
+    /// Device is busy, can't start something else
+    DeviceBusy,
+    /// Received a nack
+    Nack,
+    /// Error happened on the bus
+    BusError,
+    /// Arbitration loss,
+    ArbitrationLoss,
+    /// Overrun detected (salve mode)
+    Overrun,
+    /// Unable to compute the stop state because previous state was not expected
+    StateError,
+    /// Transfer complete status but nothing do do next
+    TransferCompleteNoRead,
+}
+
+/// State of i2c communication.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum State {
+    /// I2c is Idle
+    Idle,
+    /// Send has started
+    TxStart,
+    /// Ready for send
+    TxReady,
+    /// Data written in send buffer
+    TxSent,
+    /// Send is complete, but device is not stopped
+    TxComplete,
+    /// Send is complete
+    TxStop,
+    /// Receive has started
+    RxStart,
+    /// Ready for receive
+    RxReady,
+    /// Received is complete
+    RxStop,
+    /// Nack for send
+    TxNack,
+}
+
+impl core::fmt::Display for State {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+/// I2c1Int provides interface to communicate with i2c devices using interruptions.
+pub struct I2cInt<I2C, PINS> {
+    dev: I2c<I2C, PINS>,
+    state: State,
+    /// Last error that happened on i2c communications.
+    pub last_error: Option<I2cError>,
+    current_write_addr: Option<u8>,
+    tx_ind: usize,
+    tx_buf: Option<&'static [u8]>,
+    rx_ind: usize,
+    rx_buf: [u8; 256], // for the moment use a static buffer here.
+    recv: Option<(u8, usize)>,
+}
+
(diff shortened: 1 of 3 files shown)


src/i2cint.rs
```diff
@@ -0,0 +1,386 @@
...
+}
+
+impl core::fmt::Display for State {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+/// I2c1Int provides interface to communicate with i2c devices using interruptions.
+pub struct I2cInt<I2C, PINS> {
```
Sh3Rm4n, line 61: A `free` function is missing, so that the user can regain access to the pins, if I2c is no longer needed:

https://github.com/stm32-rs/stm32f3xx-hal/blob/9e29ec6b1dba461eb1001ea6eee2bcf255fda59f/src/i2c.rs#L201-L204

src/i2cint.rs
```diff
@@ -0,0 +1,386 @@
...
+    pub fn stop(&mut self, addr: u8) {
+        self.dev
+            .i2c
+            .cr2
+            .modify(|_, w| w.sadd().bits(u16::from(addr << 1)).stop().stop());
+    }
+
+    /// This function must be called when there is an interruption on i2c device.
+    /// It will compute the current state based on ISR register and execute work based on the state.
+    pub fn interrupt(&mut self) {
```
Sh3Rm4n, line 265: `interrupt` sound like, this function would return the type of the interrupt, which got triggered rather than reaction on the interrupt. I'd look for a more descriptive name. Maybe `react_on_interrupt()`? Better suggestions are welcome.

src/i2cint.rs
```diff
@@ -0,0 +1,386 @@
...
+
+    /// Is the device Busy ?
+    /// # Returns
+    /// true if busy, false if not.
+    pub fn is_busy(&mut self) -> bool {
+        self.dev.i2c.isr.read().busy().bit()
+    }
+
+    /// Get the state of the device.
+    pub fn get_tx_state(&self) -> State {
```
Sh3Rm4n, line 359: `State` does include RX and TX states. Therefor this function name does not make sense?

```suggestion
    pub fn get_state(&self) -> State {
```

src/i2cint.rs
```diff
@@ -0,0 +1,386 @@
...
+    /// Write a bute into the tx buffer.
+    fn write_tx_buffer(&mut self, byte: u8) {
+        self.dev.i2c.txdr.write(|w| w.txdata().bits(byte));
+        self.state = State::TxSent;
+    }
+
+    /// Is the device Busy ?
+    /// # Returns
+    /// true if busy, false if not.
+    pub fn is_busy(&mut self) -> bool {
```
Sh3Rm4n, line 354: As you already have chosen a perfect name for a boolean function, I find this description redundant:

```suggestion
    pub fn is_busy(&mut self) -> bool {
```

src/i2cint.rs
```diff
@@ -0,0 +1,386 @@
...
+    }
+
+    /// Disable the i2c device.
+    pub fn disable(&mut self) {
+        self.dev.i2c.cr1.modify(|_, w| w.pe().disabled());
+        self.last_error = None;
+    }
+
+    /// Enables all interrupts for i2c device.
+    pub fn enable_interrupts(&mut self) {
```
Sh3Rm4n, line 119: This should be more flexible in its configuration, I guess :thinking:

Sh3Rm4n (changes requested):
Thank you really much for your contribution. This looks promising. Code looks pretty clean, which I like. 

Small nit's here and there. I just have a few suggestions about the design for now. 
But I think we can improve on it further. For that, I have to read into the code a little more in detail. 

I'm not so sure, if it is a good idea to keep internal tx and rx (ring) buffers. I'd rather wish this implementation would be a thin wrapper around the `State` function, so that the user can get the received byte by themselves and put it into their preferred buffer type, like [`heapless::spsc::Queue`](https://docs.rs/heapless/0.6.1/heapless/spsc/struct.Queue.html). 

This would look something like this:

```rust
fn i2cinterrupt() {
    interrupt_type = cx.i2cinstance.get_interrupt_type();
    match interrupt_type {
        Some(TxReady) => cx.i2cinstance.read(),
        _ => ...
    }
}
```

where `get_interrupt_type()` is 

```rust
pub fn get_interrupt_type(&self) -> Option<State> { ... }
```

or even provide a function like:

```rust
pub fn read_on_rx(&self) -> Option<[u8]> {
    if self.state == State::RxReady {
        Some(self.dev.i2c.rxdr.read().rxdata().bits())
    } else {
        None
    }
}
```

Pretty much all in all a thin wrapper on the I2cInstance itself. 

Maybe we could even implement the embedded hal traits (at least `Read` and `Write` for it)

Much rambling. I hope it helped, but maybe you have better ideas :)

PR #199, opened 2021-01-29
Gpio exti

### LINKED ISSUE #112: Interrupt support

seanybaggins:
Still a work in progress.

Aims to close #112.


seanybaggins:
This code makes me sad. See #200 for my second attempt.

PR #201, opened 2021-02-05
Add documentation on how to run tests to the README.

IamfromSpace:
I encountered a bit of discovery here, so I figured I'd add some docs while it was fresh in my mind.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,3 +126,21 @@ expect it to do so.*
 ## License
 
 [0-clause BSD license](LICENSE-0BSD.txt).
+
+## Contributing
+
+### Running Tests
+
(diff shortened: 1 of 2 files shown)


IamfromSpace:
Seems that clippy is unhappy with existing code

Sh3Rm4n:
Nice, thanks for this little but useful part of documentation. It's very useful to know how to execute these tests. 

As defmt-test and the probe-run tooling improves the process might change or is getting easier. 

Clippy is failing because of a new introduced lint. 

This is fixed by https://github.com/stm32-rs/stm32f3xx-hal/pull/202

I'll rebase you branch, so that it can be merged :)

PR #202, opened 2021-02-13
Allow clippy upper_case_acronyms warnings

Piroro-hs:


seanybaggins (approved):
Had this same issue in #200 also need for #189

PR #207, opened 2021-03-07
Swap to bxCan crate

zklapow:
I said I wasn't gonna have time for this, but it turns out this was a lot easier than I realized.

Currently (as before) we only support pin defs for the 302/303, but it should be trivial to add more when thats desired.


zklapow:
Also note: this is definitely a breaking change. I don't think thats a huge deal given that can was behind a feature flag anyhow and was fairly experimental.

Sh3Rm4n:
> Also note: this is definitely a breaking change. I don't think thats a huge deal given that can was behind a feature flag anyhow and was fairly experimental.

No worries. Breaking changes for now are just important to note in the changelog. Bumping the version number is not a big deal. :)

pdgilbert:
Could you please change the version requirement to ">=0.4" or ">=0.4, <0.6" for reasons described at stm32-rs/stm32f7xx-hal#114 (comment). Discussion also available at https://github.com/stm32-rs/stm32f4xx-hal/issues/285.

zklapow:
Apologies (as usual) it took me a bit to get back to this. Should be up to date with master now with the bxcan version issue fixed, and ready for another look (or merging).

src/can.rs
```diff
@@ -1,657 +1,94 @@
...
-impl Can {
-    /// Initialize the CAN peripheral using the options specified by `opts`.
-    pub fn new_with_opts(
-        can: pac::CAN,
-        rx: gpioa::PA11<AF9<PushPull>>,
-        tx: gpioa::PA12<AF9<PushPull>>,
-        apb1: &mut APB1,
-        opts: CanOpts,
-    ) -> Can {
+    /// Note: this does not actually initialize the CAN bus. You will need to first call [`bxcan::Can::new`] and  set the bus configuration and filters before the peripheral can be enabled. See the CAN example, for a more thorough example of the full setup process.
```
Sh3Rm4n, line 74:
```suggestion
    /// Note: this does not actually initialize the CAN bus. 
    /// You will need to first call [`bxcan::Can::new`] and  set the bus configuration and filters
    /// before the peripheral can be enabled. 
    /// See the CAN example, for a more thorough example of the full setup process.
```

examples/can.rs
```diff
@@ -46,36 +48,47 @@ fn main() -> ! {
...
+        .set_bit_timing(0x001c_0003)
+        .set_loopback(false)
+        .set_silent(false);
+
+    let mut filters = can.modify_filters();
+
+    filters.enable_bank(0, Mask32::accept_all());
+
+    // Enable filters.
+    drop(filters);
```
Sh3Rm4n, line 68: This seems confusing. Has the `Drop` trait the site effect of applying the filter?

src/can.rs
```diff
@@ -1,657 +1,94 @@
...
-    pub fn free(
-        self,
-    ) -> (
-        pac::CAN,
-        gpioa::PA11<AF9<PushPull>>,
-        gpioa::PA12<AF9<PushPull>>,
-    ) {
-        (self.can, self._rx, self._tx)
-    }
-}
```
Sh3Rm4n, line 410: Hm. I see, the `bxcan` crate does consume this peripheral and is not able to free it. But I would still keep this function and submit a PR to the `bxcan` crate, that it also supports freeing the peripheral again.

src/can.rs
```diff
@@ -1,657 +1,94 @@
...
-            CanFilter::new(CanFilterData::MaskFilter(mask as u16, filter as u16))
-        }
-    }
+/// Struct representing a CAN peripheral and its configured TX and RX pins.
+///
+/// See [`bxcan::Instance`] for more information on how to use the CAN interface.
+pub struct Can<TX, RX> {
+    _can: pac::CAN,
+    _tx: TX,
+    _rx: RX,
```
Sh3Rm4n, line 64: When the `free()` method is implemented, the underscore should'nt be necessary anymore:

```suggestion
pub struct Can<TX, RX> {
    can: pac::CAN,
    tx: TX,
    rx: RX,
```

examples/can.rs
```diff
@@ -46,36 +48,47 @@ fn main() -> ! {
...
-    // Uncomment the following line to enable CAN interrupts
-    // can.listen(Event::Fifo0Fmp);
+    let mut can = bxcan::Can::new(can);
 
-    let (mut tx, mut rx0, _rx1) = can.split();
+    // Use loopback mode: No pins need to be assigned to peripheral.
+    // APB1 (PCLK1): 64MHz, Bit rate: 500kBit/s, Sample Point 87.5%
+    // Value was calculated with http://www.bittiming.can-wiki.info/
+    can.modify_config()
+        .set_bit_timing(0x001c_0003)
```
Sh3Rm4n, line 59: Minor nit: Could you give this magic value a name? I see, that your comment explains the reason for this value, but I (as someone who has not really looked into can) do not fully grasp the meaning of each single bit.

examples/can.rs
Sh3Rm4n: Ah I see, filter holds a mutual reference to can, so can can not be used later, as long as the reference is held. Is this correct? And does it still compile without this drop?

src/can.rs
```diff
@@ -1,657 +1,94 @@
...
-pub use embedded_hal_can::{self, Filter, Frame, Id, Receiver, Transmitter};
+use crate::gpio::{gpioa, gpiob};
+use crate::gpio::{PushPull, AF7, AF9};
 
-use crate::gpio::gpioa;
-use crate::gpio::{PushPull, AF9};
 use crate::pac;
 use crate::rcc::APB1;
-use nb::{self, Error};
+use bxcan::RegisterBlock;
```
Sh3Rm4n, line 19:
```suggestion
use bxcan::RegisterBlock;
pub use bxcan
```

Hm. I'm thinking about reexporting the `bxcan` crate, so one could use `bxcan` with this HAL without explicitly adding `bxcan` to the Cargo.toml itself. The suggestion is just a placeholder. I'm not sure how do to it the most clean way.

examples/can.rs
```diff
@@ -13,7 +13,9 @@ use hal::pac;
 use hal::prelude::*;
 use hal::watchdog::IndependentWatchDog;
 
-use hal::can::{Can, CanFilter, CanFrame, CanId, Filter, Frame, Receiver, Transmitter};
+use bxcan::filter::Mask32;
+use bxcan::{Frame, StandardId};
```
Sh3Rm4n, line 17: Could be 
```suggestion
use can::bxcan::{Frame, StandardId};
```
if we would reexport the can module. (See other comment)

Sh3Rm4n (changes requested):
Thanks again for your effort and no worries if it takes some time :)

Piroro-hs:
How about returning `bxcan::Can` directly from constructor?

src/can.rs
Sh3Rm4n: [bxcan 0.5.1](https://github.com/stm32-rs/bxcan/blob/master/CHANGELOG.md#051---2021-05-15) now has a `Can::free` method, so we can use it here. So bxcan instance would return `fn free(bxcan::Can<Self>) -> Self`, where Self itself implements `pub fn free(self) -> (pac::CAN, gpioa::PA11<...>, gpioa::PA12 <...>)`. 

This would mean, we have to call free twisce to get back the gpios: `let (can, pa11, pa12) = can.free().free();` but I think, it is not too bad.

Sh3Rm4n:
I've added a free method and changed some minor nits and rebased it onto current master. 

Thanks for your contribution! :)

Sh3Rm4n (approved):

Issue #177, opened 2020-12-09
Release v0.6.0


Sh3Rm4n:
- [x] Merge #173 
- [x] Merge #172
- [x] Test all examples
  - [x] adc (_only smoke test _) [1]
  - [x] can (_only smoke test_) [1]
  - [x] gpio_erased
  - [x] i2c_scanner
  - [x] pwm (_only smoke test_) [1]
  - [x] serial_dma
  - [x] spi
  - [x] toggle
  - [x] usb_serial
- [x] Run all tests
  - [x] Rcc
- [x] Update Changelog
  - [x] Top Version
  - [x] Bottom Version Link
- [x] No Dead Links
- [x] Bump Cargo.toml
- [x] Update references to version (Documentation Links)
- [ ] Tag and Relase on Github
- [ ] Release on crates.io

[1]: Did not panic!

PR #212, opened 2021-03-28
Add more tests

Sh3Rm4n:
With focus on gpio tests for now. This is very much WIP. 

I'd like to release a new version of this crate soon, but I'd also like to reduce the smoke-testing work (which I've done for 0.6.0 in https://github.com/stm32-rs/stm32f3xx-hal/issues/177) while automating some testing. 

- [x] fix open drain test
- [x] add serial test
- [ ] add spi test
- [ ] add i2c test (if it is feasible, because slave is not implemented IIRC)
- [x] make sure testsuite is build on CI


Sh3Rm4n:
SPI and I2C will have to wait for later PRs. This includes slight changes to watchdog and adc, which I did not expect. So to not blow this PR up anymore, I'll keep it as is

Issue #215, opened 2021-04-19
Getting it to run on STM32F303, can't compile


lucaszanella:
I'm very noob at this, would appreciate some help.

I'm trying the example https://github.com/stm32-rs/stm32f3xx-hal/blob/master/examples/pwm.rs on my STM32F303 Discovery board. Here's my cargo:

```
[package]
name = "stmf32f303_tests"
version = "0.1.0"
authors = ["Lucas Zanella <me@lucaszanella.com>"]
edition = "2018"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
embedded-hal = "0.2.3"
nb = "0.1.2"
cortex-m = "0.6.2"
cortex-m-rt = "0.6.11"
# Panic behaviour, see https://crates.io/keywords/panic-impl for alternatives
panic-halt = "0.2.0"

[dependencies.stm32f3xx-hal]
version = "0.6.1"
features = ["rt", "stm32f303xc", "defmt"]
```

I've put the code of `pwm.rs` into `main.rs`. 

Errors:

```
error[E0599]: no method named `MHz` found for type `u32` in the current scope
  --> src/main.rs:31:40
   |
31 |     let clocks = rcc.cfgr.sysclk(16u32.MHz()).freeze(&mut flash.acr);
   |                                        ^^^ method not found in `u32`

error[E0599]: no method named `into_af2_push_pull` found for struct `stm32f3xx_hal::gpio::gpioa::PA4<Input<Floating>>` in the current scope
  --> src/main.rs:37:10
   |
37 |         .into_af2_push_pull(&mut gpioa.moder, &mut gpioa.otyper, &mut gpioa.afrl);
   |          ^^^^^^^^^^^^^^^^^^ method not found in `stm32f3xx_hal::gpio::gpioa::PA4<Input<Floating>>`

error[E0599]: no method named `into_af2_push_pull` found for struct `stm32f3xx_hal::gpio::gpioa::PA6<Input<Floating>>` in the current scope
  --> src/main.rs:40:10
   |
40 |         .into_af2_push_pull(&mut gpioa.moder, &mut gpioa.otyper, &mut gpioa.afrl);
   |          ^^^^^^^^^^^^^^^^^^ method not found in `stm32f3xx_hal::gpio::gpioa::PA6<Input<Floating>>`

[...]

PR #216, opened 2021-04-21
Auto generate memory.x

Piroro-hs:
This PR adds `ld` feature, which enables the `memory.x` generation and prevent issues like #215.


build.rs
```diff
@@ -0,0 +1,65 @@
...
+    if cfg!(feature = "ld") {
+        gen_memory_x();
+    }
+    println!("cargo:rerun-if-changed=build.rs");
+}
+
+fn gen_memory_x() {
+    #![allow(clippy::unneeded_wildcard_pattern)]
+
+    let mem_cfg_set = (
```
Sh3Rm4n, line 13: Hm. Isn't an enum a better option here? Then the following code should be much more readable.

src/lib.rs
```diff
@@ -69,26 +74,31 @@ compile_error!(
...
-    *   stm32f303xe
-    *   stm32f303x6
-    *   stm32f303x8
-    *   stm32f373
-    *   stm32f378
-    *   stm32f334
-    *   stm32f328
-    *   stm32f358
-    *   stm32f398
+    * stm32f301x6
```
Sh3Rm4n, line 77: This list gets longer and longer. I think we should reduce this error message to the essential minimum and point to the documentation, what devices / features are supported.

.github/workflows/ci.yml
```diff
@@ -12,30 +12,34 @@ jobs:
     strategy:
       matrix:
         mcu:
-          - stm32f301
-          - stm32f318
+          - stm32f301x6
+          - stm32f318x8
+          - stm32f302x6
+          # - stm32f302x8
```
Sh3Rm4n, line 18: Why are these currently commented out?

Cargo.toml
```diff
@@ -66,9 +66,24 @@ default = ["unproven"]
...
 rt = ["stm32f3/rt"]
 can = ["embedded-hal-can"]
 
+svd-f301 = ["stm32f3/stm32f301"]
+svd-f302 = ["stm32f3/stm32f302"]
+svd-f303 = ["stm32f3/stm32f303"]
+svd-f373 = ["stm32f3/stm32f373"]
+svd-f3x4 = ["stm32f3/stm32f3x4"]
+
+mem-4 = []
```
Sh3Rm4n, line 79: Out of curiosity, where to have the namings from? Can you point me towards the according documentation?

build.rs
```diff
@@ -0,0 +1,65 @@
...
+        cfg!(feature = "mem-4"),
+        cfg!(feature = "mem-6"),
+        cfg!(feature = "mem-8"),
+        cfg!(feature = "mem-b"),
+        cfg!(feature = "mem-c"),
+        cfg!(feature = "mem-d"),
+        cfg!(feature = "mem-e"),
+    );
+    let flash = match mem_cfg_set {
```
Sh3Rm4n, line 22: It would be nice to add a comment, where you found these numbers.

src/lib.rs
Sh3Rm4n: Note: The `TODO` in line 66 is no longer true. I tripped up on my own TODO note. If you like to, you can remove it :)

src/lib.rs
```diff
@@ -97,91 +107,52 @@ compile_error!(
...
-    ///
-    /// Uses core function, when defmt is not active
-    #[macro_export]
-    macro_rules! unwrap {
-        ($l:expr) => {
-            $l.unwrap()
-        };
-    }
+cfg_if! {
+    if #[cfg(feature = "device-selected")] {
```
Sh3Rm4n, line 113: Because missing `device-selected` does throw a `compile_error!` above already, I wonder if this check is needed at all. 
You could remove it, to save us some indentation :wink:

Sh3Rm4n (changes requested):
I like, that the current `memory.x` file is removed now. It was only added to be able to build examples, but was not intended to be used by the user. Having the `ld` feature for an opt in **correct** `memory.x` files is the correct way to go, IMO :+1: 

I guess this feature is worth a mention in the CHANGELOG.md? :)

src/lib.rs
Rahix: I suggest gating everything on `device-selected` anyway.  This way, if no device is selected, the only error message is the one from `compile_error!` and there aren't millions of "item not found" errors alongside it which clutter the compiler output and make it hard to see the original error.

src/lib.rs
Rahix: Maybe this message can be compacted without dropping the convenience of listing all supported devices by putting multiple names on a single line?  For example:

```rust
compile_error!(
    "This crate requires you to specify your target chip as a feature.

    Please select one of the following (`x` denotes any character in [a-z]):

        stm32f301x6 stm32f301x8
        stm32f318x8
        stm32f302x6 stm32f302x8 stm32f302xb stm32f302xc stm32f302xd stm32f302xe
        stm32f303x6 stm32f303x8 stm32f303xb stm32f303xc stm32f303xd stm32f303xe
        stm32f328x8
        stm32f358xc
        stm32f398xe
        stm32f373x8 stm32f373xb stm32f373xc
        stm32f378xc
        stm32f334x4 stm32f334x6 stm32f334x8

    Example: The STM32F3Discovery board has a STM32F303VCT6 chip.
    So you want to expand your call to `cargo` with `--features stm32f303xc`.
    For more information, see README -> Selecting the right chip.
    "
);
```

Sh3Rm4n: That would be a good alternative. I'm fine with both.

src/lib.rs
Sh3Rm4n: Good point. Than keep it as is :)

Cargo.toml
Piroro-hs: Well, 4/6/8/b/c/d/e is just came from part number

src/lib.rs
Piroro-hs: I found that we should check `device-selected` in `build.rs`, to avoid `thread 'main' panicked at 'internal error: entered unreachable code', build.rs:39:9` when enable `ld` feature without specifying device (`cargo check --features ld`).

Cargo.toml
Sh3Rm4n: I see 😄

src/lib.rs
Sh3Rm4n: Either that, or throwing an error, which gives a clear hint to the cause. 

But checking if `device-selected` was set at the before calliong`gen_memory_x` makes sense. Anyway, I would still prefer clearer error-handling instead of `unreachable!` in the `_`-arm in the match cases.

build.rs
```diff
@@ -0,0 +1,65 @@
+use std::{env, fs::File, io::prelude::*, path::PathBuf};
+
+fn main() {
+    if cfg!(feature = "ld") {
```
Sh3Rm4n, line 4:
```suggestion
    if cfg!(feature = "ld") and cfg!(feature = "device-selected" {
```

as discussed in https://github.com/stm32-rs/stm32f3xx-hal/pull/216#discussion_r620266515

build.rs
```diff
@@ -0,0 +1,65 @@
...
+    );
+    let flash = match mem_cfg_set {
+        (true, ..) => 16,
+        (_, true, ..) => 32,
+        (_, _, true, ..) => 64,
+        (.., true, _, _, _) => 128,
+        (.., _, true, _, _) => 256,
+        (.., _, _, true, _) => 384,
+        (.., _, _, _, true) => 512,
+        _ => unreachable!(),
```
Sh3Rm4n, line 30: Replace with cleaner error handling, as discussed in https://github.com/stm32-rs/stm32f3xx-hal/pull/216#discussion_r620266515

build.rs
```diff
@@ -0,0 +1,124 @@
...
+    } else if cfg!(feature = "mem-b") {
+        Mem::B
+    } else if cfg!(feature = "mem-c") {
+        Mem::C
+    } else if cfg!(feature = "mem-d") {
+        Mem::D
+    } else if cfg!(feature = "mem-e") {
+        Mem::E
+    } else {
+        unreachable!()
```
Sh3Rm4n, line 80: Even though this is clearly unreachable for the current state of the file, I'd prefer an error message + abort / exit(1), which would be much more helpful for the user, if this line is somehow executed in the future.

src/lib.rs
```diff
@@ -2,100 +2,76 @@
...
+ * stm32f373x8
+ * stm32f373xb
+ * stm32f373xc
+ * stm32f378xc
+ * stm32f334x4
+ * stm32f334x6
+ * stm32f334x8
+
+ Example: The STM32F3Discovery board has a STM32F303VCT6 chip.
+ So you need to specify `stm32f303xc` in your `cargo.toml` (note that VC → xc).
```
Sh3Rm4n, line 44:
```suggestion
 So you need to specify `stm32f303xc` in your `Cargo.toml` (note that VC → xc).
```

src/lib.rs
```diff
@@ -2,100 +2,76 @@
...
+
+ For more information, see the [README][].
+
+ [README]: https://github.com/stm32-rs/stm32f3xx-hal/blob/master/README.md#selecting-the-right-chip
+
+ ### ld
```
Sh3Rm4n, line 50: Pretty minor nit (can be ignored) I prefer an empty line after a section header.

Sh3Rm4n (commented):
Very much appreciate the documentation updates and moving the device-selected check towards the build.rs file is a good solution as well! :) 

Nice work. Resolving the commented out devices in CI and after that, this looks good to be merged.

src/lib.rs
Piroro-hs: Oops

.github/workflows/ci.yml
Piroro-hs: Devices only differ in memory size are commented out as they only check the literally same code.

Sh3Rm4n: Hm. This does make sense. Good catch :+1: Could you at a comment, so this is obvious, when editing the file, so this will not be accidentally deleted? :)

Sh3Rm4n:
I was thinking about maybe adding the `ld` feature to the default features, as well as the `rt` feature. This should be much more user friendly initially and experienced user can turn of default features anyways. 

But I think, this is a discussion for later. Let's first release a version with `memory.x` auto-generation and wait for feedback. Just a note for the future.

Sh3Rm4n (approved):
LGTM. If you don't say otherwise, I would merge it :)

Issue #206, opened 2021-03-06
adc delays for seconds on init


David-OConnor:
This is run after calibrate:

adc.rs:
```rust
// ADEN bit cannot be set during ADCAL=1
// and 4 ADC clock cycle after the ADCAL
// bit is cleared by hardware

// ...

fn wait_adc_clk_cycles(&self, cycles: u32) {
    let adc_clk_cycle = self.clocks.hclk().0 / (self.ckmode as u32);
    asm::delay(adc_clk_cycle * cycles);
}
```

I'm not sure what an ADC clock cycle is, but don't think it's right here. For example, the first variable will be on the order of 10^6. I think maybe it should be `asm::delay(self.ckmode as u32 * cycles);` And, remove this fn and move the delay to right after the cal, since it's now a one-line one-off.

This causes consequences like long boot-up time, and disconnecting USB devices (or anything that requires a quick response and is initialized before the ADC)

Sh3Rm4n:
Currently, I don't really see the concrete problem / best solution (have'nt looked deeply into it), but this is clearly a problem of being a blocking function. 

I think this function should not be called implicitly inside the code and should be visible to the user. At best, this function could just calculate an estimate time, which is needed, so that the ADC is guaranteet to be ready (calibration is done) and the user can choose the way to go about it (blocking himself via `asm::delay` or some timer, or going the async / dispension point route)

David-OConnor:
Is the original code consistent with the RM note about cycles? Ie is a multi-second delay what's specified, or only a multi-tick delay?

Sh3Rm4n:
I don't know. I would have to look into the RM myself, to find out.

David-OConnor:
The main part is what you put in the comment. I'm just not sure how to interpret it. Leaning towards the modification I posted here.

Sh3Rm4n:
Well in your suggestions the delay is dependent on the system clock, which is in most cases different from the clock, which the adc peripheral uses. So this could lead to problems or oversights. Nonetheless, the current implementation has non-desirable behavior. 

I don't know if the delay calculation is wrong and does lead to too long delays, or if we need a non-blocking way.

David-OConnor:
I give up

Sh3Rm4n:
I'm sorry that you are unhappy with the responsiveness. 

But I can't help you but say that this is an open source project managed by volunteers which devote their free time into this project. 

Currently I'm the only (not so) frquent maintainer of this repository. 
[...]

PR #217, opened 2021-04-25
Fix long delay during ADC initialization

Rahix:
As explained in issue #206, the code here is doing the calculation completely wrong.  This leads to a multi-second startup delay instead of just a few clock cycles.

Fix this by using the correct calculation; currently only ADC clock configurations synchronous to the AHB are supported which makes this quite easy: We just need to delay the given number of cycles * the selected ADC clock prescaler (1, 2, or 4).

To make this code future proof against an implementation of asynchronous ADC clock configurations, use a `match` statement here which will fail to compile once the `ASYNCHRONOUS` variant is added to the `CkMode` struct - thus forcing the implementor to take another look at this.

Fixes: #206

### DIFF
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -399,8 +399,15 @@ macro_rules! adc_hal {
                 }
 
                 fn wait_adc_clk_cycles(&self, cycles: u32) {
-                    let adc_clk_cycle = self.clocks.hclk().0 / (self.ckmode as u32);
-                    asm::delay(adc_clk_cycle * cycles);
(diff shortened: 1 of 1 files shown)


Rahix:
(FYI, you manually need to approve the CI run now due to [recent changes with GHA](https://github.blog/2021-04-22-github-actions-update-helping-maintainers-combat-bad-actors/))

Sh3Rm4n:
Thanks! Very much appreciated

PR #220, opened 2021-04-29
Improve build.rs and some more. 

Sh3Rm4n:
I had some ideas about the newly added build.rs and touched some documentation as well. 

What do you think about this change, @Piroro-hs?

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,4 +1,14 @@
-use std::{env, fs::File, io::prelude::*, path::PathBuf};
+use std::collections::hash_set::HashSet;
+use std::env;
+use std::fs::File;
+use std::io::prelude::*;
+use std::path::PathBuf;
+
+use cargo_metadata::MetadataCommand;
+// TODO(Sh3Rm4n - 2021-04-28):
+// Remove when [feature="slice_group_by"] is stable.
+// See with https://github.com/rust-lang/rust/issues/80552.
+use slice_group_by::GroupBy;
 
 fn main() {
     check_device_feature();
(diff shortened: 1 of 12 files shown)


build.rs
```diff
@@ -10,40 +19,69 @@ fn main() {
...
-Please select one of the following (`x` denotes any character in [a-z]):
+    // get all device variants from the metadata
+    let metadata = MetadataCommand::new().exec().unwrap();
+    let mut device_variants = metadata.root_package().unwrap().features.clone();
+    device_variants.retain(|_, dependend_features| {
+        dependend_features
+            .iter()
+            .any(|feature| *feature == "device-selected")
+    });
+    let mut device_variants = device_variants.keys().cloned().collect::<Vec<String>>();
```
Piroro-hs, line 44: One `filter_map` call is enough?

Piroro-hs (changes requested):
Ensuring user select exactly one device is great I think :)

build.rs
Piroro-hs:
```rust
    let features = metadata.root_package().unwrap().features.iter();
    let mut device_variants: HashSet<String> = features
        .filter_map(|(feature, dependend_features)| {
            dependend_features
                .iter()
                .any(|dependend_feature| dependend_feature == "device-selected")
                .then(|| feature.clone())
            // `bool::then` is only available for Rust >= 1.50
            // match dependend_features
            //     .iter()
            //     .any(|feature| feature == "device-selected")
            // {
            //     true => Some(feature.clone()),
            //     false => None,
            // }
        })
        .collect();
```

Piroro-hs:
How about changing those `eprintln!`s like this?
```rust
    eprintln!(
        "I've just found Rust's string literal support line breaking.\n\
        \n\
        We can keep indentation!!!."
    );
```

Sh3Rm4n:
Oh, I've searched for a variation to keep the indentation as well. :smile: Thanks for the tip  :+1:

build.rs
```diff
@@ -10,40 +19,69 @@ fn main() {
...
+        .collect();
 
-Example: The STM32F3Discovery board has a STM32F303VCT6 chip.
+    // check if exactly one device was selected
+    if device_variants
+        .iter()
+        .filter(|device| selected_features.contains(&device.to_ascii_uppercase()))
+        .count()
+        != 1
+    {
```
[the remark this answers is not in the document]
Sh3Rm4n: I really like this idea. I should look more into the `Set` types.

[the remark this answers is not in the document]
Sh3Rm4n: Hm. I think I'll go for a [`Union`](https://doc.rust-lang.org/std/collections/hash_set/struct.Union.html) instead. It's essentially the same, but more obvious. I had to look up, what the result of a `BitOr` was between to `HashSet` even though it's pretty clear in hindsight.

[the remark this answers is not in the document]
Sh3Rm4n: [`intersection`](https://doc.rust-lang.org/std/collections/hash_set/struct.HashSet.html#method.intersection) is what I was wanting to go for.

build.rs
Sh3Rm4n: Bumping MSVR is not a big deal, so I'll go for the `bool::then` method.

build.rs
```diff
@@ -10,40 +20,75 @@ fn main() {
...
+    if !cfg!(feature = "device-selected") && cfg!(feature = "direct-call-deprecated") {
+        eprintln!(
+            "The (device-)feature you selected is deprecated, because it was split up into sub-devices.\n\
+            \n\
+            Example: The STM32F3Discovery board has a STM32F303VCT6 chip.\n\
+            You probably used to use `stm32f303` but now functionalities for the sub-device were added.\n\
+            In this case replace it with `stm32f303xc` to make your code build again.\n\
+            \n\
+            For more information, see \
+            \x1b]8;;https://github.com/stm32-rs/stm32f3xx-hal#selecting-the-right-chip\x1b\\README\
```
Piroro-hs, line 33:
```suggestion
            \x1b]8;;https://github.com/stm32-rs/stm32f3xx-hal#selecting-the-right-chip\x1b\\README \
```

build.rs
```diff
@@ -10,40 +20,75 @@ fn main() {
...
-You probably used to use `stm32f303` but now functionalities for the sub-device were added.
-In this case replace it with `stm32f303xc` to make your code build again.
+    // get all device variants from the metadata
+    let metadata = MetadataCommand::new().exec().unwrap();
+    let device_variants: HashSet<String> = metadata
+        .root_package()
+        .unwrap()
+        .features
+        .iter()
+        .filter_map(|(feature, dependend_features)| {
```
Piroro-hs, line 46:
```suggestion
        .filter_map(|(feature, dependent_features)| {
```

Piroro-hs (commented):
Found some typo

Sh3Rm4n:
Thanks for your review @Piroro-hs

PR #224, opened 2021-05-19
MVP: Make use of const generics

Sh3Rm4n:
This is a MVP of using const generics, based on the refactoring of the GPIO module in https://github.com/stm32-rs/stm32f3xx-hal/pull/189

The current implementation leaves out the `downgrade()` method to get a generic pin type, which is a major loss. 
This is because we can not be generic over `Unsigned` via min const generics. 
 min const generic does not allow any const generics in `where` clauses, and I'm not really sure, if this would be the right approach even if we had full const generics available. 
 
 Maybe there are some improvements to be made, which would allow not having to remove this method.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ exclude = [
     "codegen",
     ".markdownlint.yml"
 ]
+resolver = "2"
 
 [package.metadata.docs.rs]
 features = ["stm32f303xc", "rt", "stm32-usbd", "can"]
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -49,31 +49,31 @@ pub struct Adc<ADC> {
 ///
 /// Each channel can be sampled with a different sample time.
 /// There is always an overhead of 13 ADC clock cycles.
-/// E.g. For Sampletime T_19 the total conversion time (in ADC clock cycles) is
+/// E.g. For Sampletime T19 the total conversion time (in ADC clock cycles) is
 /// 13 + 19 = 32 ADC Clock Cycles
 pub enum SampleTime {
     /// 1.5 ADC clock cycles
-    T_1,
+    T1,
     /// 2.5 ADC clock cycles
-    T_2,
+    T2,
     /// 4.5 ADC clock cycles
-    T_4,
+    T4,
     /// 7.5 ADC clock cycles
-    T_7,
+    T7,
     /// 19.5 ADC clock cycles
-    T_19,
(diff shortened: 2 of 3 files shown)


Sh3Rm4n:
Hey @Piroro-hs, I would like to ask you for your help. 

The initial motivation of the PR making use of const generics was the gnarly error message introduced by #189, like
this one https://github.com/stm32-rs/stm32f3xx-hal/runs/2623986440#step:4:131. 

I stumbled upon these error messages, while implementing test cases in https://github.com/stm32-rs/stm32f3xx-hal/pull/212, 
and I would rather not release a version until this issue is fixed. 

I know that `typeenum` is a pretty powerful tool to be generic over integers, but I feel like the current error messages are a major loss, taking into consideration, that this HAL implementation is strongly typed partly because the error messages can give such a could hint about a mistake. 

I don't know what to do though. Reverting #189 would be the easiest way, but your great improvement are worthwhile, and I would like to keep them. 
I was thinking a lot about an elegant solution, taking the approach in this PR, but I have not come to a conclusion, maybe you have some insights. 

The easiest way I was thinking about, is adding a new generic type PXx (which is distinct from the current generic Pin) type, which has no generic parameter about the pin number. This would result into many duplications though, as we have to implement many methods again for this pin (which would reintroduce some macro bulk which you have been getting rid of in your improvment PR)

Do you have any idea or suggestion @Piroro-hs ?

Piroro-hs:
My idea: https://github.com/Piroro-hs/stm32f3xx-hal/tree/untypenum

Main idea is that using this simple struct instead of `typenum` crate
```rust
pub struct U<const X: u8>;
```

Sh3Rm4n:
Well this is a much simpler solution than mine :grinning: And it solves the problem the most prominent problem of typeenum error types. `typenum::Unsigned` as a trait is still pretty useful, and I don't see, that it can be replaced with `min-const-generics`. 

Let's go your route for now. Switching to `const-generics` would be a dream, but let's wait until that is more fleshed out.

What I still might suggest in incorporate  https://github.com/stm32-rs/stm32f3xx-hal/pull/224/commits/1ef350ae73443a18723559b7dda4e1566406d1bb  with your (https://github.com/Piroro-hs/stm32f3xx-hal/commit/b34576ba0a14b296f7b953d2147566cfe403d4ce) change, so that we can switch to `const-generics` while the constraint still remains, that const generics have to be in the last position.

PR #227, opened 2021-05-25
Make use of const generics (take 2)

Piroro-hs:
Closes #224


Sh3Rm4n:
This is much better :tada:  thank you :) 

```txt

error[E0599]: the method `into_af7_push_pull` exists for struct `stm32f3xx_hal::gpio::Pin<Gpiob, U<9_u8>, Input>
`, but its trait bounds were not satisfied
   --> testsuite/tests/uart.rs:65:18
    |
65  |                 .into_af7_push_pull(&mut gpiob.moder, &mut gpiob.otyper, &mut gpiob.afrh),
    |                  ^^^^^^^^^^^^^^^^^^
    |
   ::: /home/fabian/Projects/stm32f3xx-hal/src/gpio.rs:272:1
    |
272 | pub struct Pin<Gpio, Index, Mode> {
    | --------------------------------- doesn't satisfy `_: IntoAf7`
    |
    = note: the following trait bounds were not satisfied:
            `stm32f3xx_hal::gpio::Pin<Gpiob, U<9_u8>, Input>: IntoAf7`

```

I find the error message still a bit leaky (leaking irrelevant implementation details of the gpio module), which is partially due to the type alias used.  But I guess, we have to live with it, as it is not easily solvable without sacrificing flexibility in the generic implementation. Maybe the error message will improve with later rustc versions :grin:

PR #232, opened 2021-06-07
Add read and write impl for Serial, Irq flags functions and RTIC example

brain113:
Hi,
This pull request adds Read and Write implementations to Serial and few functions to work with Irq flags.
This needed to add more control under communicaiton flow and timing.

Also I added an example how to use this with RTIC. The example uses DIR pin when transmiting (like RS485).

PS: This is my first pull request ever :)

### DIFF
--- a/examples/serial_echo_rtic.rs
+++ b/examples/serial_echo_rtic.rs
@@ -7,14 +7,23 @@ use panic_rtt_target as _;
 mod app {
     use dwt_systick_monotonic::DwtSystick;
     use rtt_target::{rprintln, rtt_init_print};
-    use stm32f3xx_hal::{prelude::*, serial::{Event, Serial}};
-    use stm32f3xx_hal::gpio::{self, Output, PushPull, Alternate, U};
+    use stm32f3xx_hal::{
+        gpio::{self, Output, PushPull, AF7},
+        prelude::*,
+        serial::{Event, Serial},
+    };
 
     #[monotonic(binds = SysTick, default = true)]
     type DwtMono = DwtSystick<48_000_000>;
 
-    type SerialType = Serial<stm32f3xx_hal::pac::USART1, (gpio::Pin<gpio::Gpioa, U<9_u8>, Alternate<PushPull, 7_u8>>, gpio::Pin<gpio::Gpioa, U<10_u8>, Alternate<PushPull, 7_u8>>)>;
-    type DirType = stm32f3xx_hal::gpio::Pin<gpio::Gpioe, U<13_u8>, Output<PushPull>>;
+    type SerialType = Serial<
+        stm32f3xx_hal::pac::USART1,
+        (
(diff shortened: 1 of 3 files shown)


examples/serial_echo_rtic.rs
```diff
@@ -0,0 +1,106 @@
...
+    use dwt_systick_monotonic::DwtSystick;
+    use rtt_target::{rprintln, rtt_init_print};
+    use stm32f3xx_hal::{prelude::*, serial::{Event, Serial}};
+    use stm32f3xx_hal::gpio::{self, Output, PushPull, Alternate, U};
+
+    #[monotonic(binds = SysTick, default = true)]
+    type DwtMono = DwtSystick<48_000_000>;
+
+    type SerialType = Serial<stm32f3xx_hal::pac::USART1, (gpio::Pin<gpio::Gpioa, U<9_u8>, Alternate<PushPull, 7_u8>>, gpio::Pin<gpio::Gpioa, U<10_u8>, Alternate<PushPull, 7_u8>>)>;
+    type DirType = stm32f3xx_hal::gpio::Pin<gpio::Gpioe, U<13_u8>, Output<PushPull>>;
```
Sh3Rm4n, line 17: Even though these are the correct types hinted by the compiler, there exists type aliases, which are easier to read. I'd recommend them: 

```suggestion
    type SerialType = Serial<stm32f3xx_hal::pac::USART1, (gpio::gpioa::PA9<A7<PushPull>>, gpio::gpioa::PA10<A7<PushPull>>)>;
    type DirType = stm32f3xx_hal::gpio::gpioe::PE13<Output<PushPull>>;
```

Sh3Rm4n (changes requested):
> PS: This is my first pull request ever :)

Whoooho. Pretty exciting stuff. I'm happy to be the reviewer of your first PR 🎉 

This PR looks pretty good. One minor nit, and you also have to make rustfmt happy via `cargo fmt --all` and `clippy` has also some minor thing to say: https://github.com/stm32-rs/stm32f3xx-hal/runs/2765167229

Thanks for your PR. I'm pretty excited about this as well, as this comes with the first RTIC example for this project! 😄

brain113:
Hi,
thank you for the warm welcome! 
I added additional commit to make rustfmt and clippy happy.

Actually I don't understand why clippy make me to change 
`init::LateResources {serial, dir}`
to 
`init::LateResources {dir, serial}`
the struct definition is 
`    struct Resources {
        serial: SerialType,
        dir: DirType,
    }`
so the "serial" field is the first.

Sh3Rm4n:
This is pretty strange indeed. A wild guess: Maybe it has something to do with the rtic `#[resources]` proc-macro, which reorders the fields under the hood and this makes clippy complain? 
I don't know. Would be worth looking into, but it does not really matter right now. 

Thank you really much for your contribution. Nice PR! :+1:

Sh3Rm4n (approved):

PR #237, opened 2021-06-12
Fix off-by-one error in Delay impl

jamesmunns:
This makes two changes:

* Updates the MAX_RVR to a 24-bit (not 25-bit) number
* Moves tracking variable update for better potential precision

This brings the stm32f3 delay impl up to date with the [nrf52 delay impl](https://github.com/nrf-rs/nrf-hal/blob/master/nrf-hal-common/src/delay.rs#L46-L72).

### DIFF
--- a/src/delay.rs
+++ b/src/delay.rs
@@ -64,8 +64,8 @@ impl DelayMs<u8> for Delay {
 
 impl DelayUs<u32> for Delay {
     fn delay_us(&mut self, us: u32) {
-        // The RVR register is 24 bits wide, as SysTick is based on a 24 bit counter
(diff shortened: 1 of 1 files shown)


sirhcel:
@Sh3Rm4n, this change works a treat on top of my https://github.com/sirhcel/stm32f3xx-hal/commit/2a9b8f7e897b81bdaf9228550a01f8564709fda3. Isn't this something for the next release as well?

Sh3Rm4n:
Thanks @jamesmunns! This for fixing this bug and keeping this HAL implementation up to date with the other implementations. 

Only rustfmt is unhappy, so this PR is almost good to go. I can fix this myself, if you don't mind though :)

PR #238, opened 2021-06-16
Allow Serial::new instead of Serial::usartX

Sh3Rm4n:
This requires changing I2C to a more generic implementation,
where the specific UART peripheral implementation get's dispatch
upon it's Instance trait.

The specific UART is longer hard coded into, but instead is referenced
via pointer (trait).

This also allows to reduce the usart macro implementation, so that
the code is easier to follow.

But this leads to code duplication, as code for the Write and Read
trait of embedded-hal can no longer be shared between Serial and Tx |
Rx.

Also, the underlying UART peripheral of Tx and Rx after split has to be
tracked at runtime, instead of being decoded at compile time (via macro).

In addition to all of this the Idle interrupt event was also added.


Sh3Rm4n:
Tested via
- [x] `cargo test -p testsuite` 
- [x] `cargo test-p testsuite --relase`

PR #250, opened 2021-07-06
Make PXx structs `Send` and `Sync`

### LINKED ISSUE #249: Should `Gpiox` be marked `Send`?
I'm using the erased `PXx` types so I can store pins from multiple ports into a single array.  I noticed `Gpiox` uses a pointer, so it does not implement `Send`.  However, it looks like we only ever do atomic or side-effect free operations on this pointer - does that mean `Send` (or even `Sync`) would be appropriate to implement?  

I'm not enough of an expert on unsafe rust to know for sure, but according to my reading of [the rustonomicon](https://doc.rust-lang.org/nomicon/send-and-sync.html) this would be allowed.

russellmcc:
Closes #249

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -133,6 +133,12 @@ macro_rules! gpio {
             _mode: PhantomData<MODE>,
         }
 
+        // NOTE(unsafe) we only use atomic operations on the gpio registers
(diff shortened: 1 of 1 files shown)


russellmcc:
oops, committed on a really old checkout.  will re-open momentarily, sorry for the noise.

PR #251, opened 2021-07-06
Mark Gpiox as Send and Sync

### LINKED ISSUE #249: Should `Gpiox` be marked `Send`?

russellmcc:
This is appropriate because we only ever perform atomic operations on
these registers.  This allows `PXx` to be Send + Sync.

Closes #249

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Implement `Serial::join` which allows to re-create the serial peripheral,
   when `Serial::split` was previously called. ([#252])
 
-## Breaking Changes
+### Changed
+
+- `PXx` struct (representing a generic GPIO pin) implements `Send` and `Sync` [#251]
+
+### Breaking Changes
 
 - Refactor CAN to use the [`bxCan`](https://github.com/stm32-rs/bxcan) crate. ([#207])
 
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -183,6 +183,18 @@ pub struct Gpiox {
     index: u8,
 }
 
+// # SAFETY
+// As Gpiox uses `dyn GpioRegExt` pointer internally, `Send` is not auto-implemented.
+// But since GpioExt does only do atomic operations without side-effects we can assume
+// that it safe to `Send` this type.
+unsafe impl Send for Gpiox {}
+
+// # SAFETY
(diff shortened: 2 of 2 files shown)


src/gpio.rs
```diff
@@ -183,6 +183,12 @@ pub struct Gpiox {
     index: u8,
 }
 
+// NOTE(unsafe) we only use atomic operations on the gpio registers
```
Sh3Rm4n, line 186: Hm. Maybe a more informative comment could be more clear. Something in the lines of

```suggestion
// # SAFETY
// As Gpiox uses dyn GpioExt internally, `Send` is not auto-implement. 
// But since GpioExt does only do atomic operations without side-effects we can assume, 
// that it safe to `Send` this type. 
```

Sh3Rm4n (changes requested):
Nice. Thanks for catching that `Gpiox` is not `Send` and `Sync`. I have some minor suggestions. And as this affects the API I would be happy about a changelog entry :)

CHANGELOG.md
```diff
@@ -13,7 +13,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Basic serial implementation also available for UART4 and UART5 [#246]
 - Implement serial DMA also for Serial [#246]
 
-## Breaking Changes
+### Changed
```
Sh3Rm4n, line 16: Minor. Markdown lint wants spaces after headings

```suggestion

### Changed
```

Sh3Rm4n (approved):
Thanks! LGTM. There are unrelated CI failures which I'm about to fix in #252 so you can ignore these. 

Either you rebase when it's merged or I will merge ignoring the failing CI :)

russellmcc:
Thanks for the review!  No preference on my end between you merging this with CI failures or rebasing after the CI fix.  Let me know.

Sh3Rm4n:
Let's rebase then :)

PR #253, opened 2021-07-09
Rework serial (interrupt) events 

### LINKED ISSUE #245: Allow more interrupt sources for Serial
Here is a screenshot out of RM0316

image

Sh3Rm4n:
Closes #245

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,7 @@ stm32f3 = "0.13.2"
 bxcan = { version = ">=0.4, <0.6", optional = true }
 stm32-usbd = { version = "0.6.0", optional = true }
 void = { version = "1.0.2", default-features = false }
+enumset = "1.0.6"
 
 [dev-dependencies]
 cortex-m = "0.7.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ stm32f3 = "0.13.2"
 bxcan = { version = ">=0.4, <0.6", optional = true }
 stm32-usbd = { version = "0.6.0", optional = true }
 void = { version = "1.0.2", default-features = false }
-enumset = "1.0.6"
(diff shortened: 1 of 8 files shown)


src/serial.rs
```diff
@@ -188,6 +246,8 @@ mod split {
 
 pub use split::{Rx, Tx};
 
+type StopModeWakeup = pac::usart1::cr3::WUS_A;
```
Sh3Rm4n, line 249:
```suggestion
pub type StopModeWakeup = pac::usart1::cr3::WUS_A;
```
?

src/timer.rs
```diff
@@ -202,17 +209,10 @@ macro_rules! hal {
                     timer
                 }
 
-                /// Starts listening for an `event`
-                pub fn listen(&mut self, event: Event) {
+                /// Enable or disable the interrupt for the specified [`Event`].
+                pub fn configure_interrupt(&mut self, event: Event, enable: bool) {
```
Sh3Rm4n, line 213: Maybe add a return value, which signals, if the interrupt was enabled or not, because some `Event`s (which are not added yet) can not be used for interrupts.

Cargo.toml
```diff
@@ -47,6 +47,7 @@ stm32f3 = "0.13.2"
 bxcan = { version = ">=0.4, <0.6", optional = true }
 stm32-usbd = { version = "0.6.0", optional = true }
 void = { version = "1.0.2", default-features = false }
+enumset = { version = "1.0.6", optional = true}
```
Sh3Rm4n, line 50: Reminder to self: Do not forget to mention this feature in 

https://github.com/stm32-rs/stm32f3xx-hal/blob/598fb342c4d64c1b13d5d1147d9e50a236739bb6/src/lib.rs#L50

PR #255, opened 2021-07-13
Support USB on the stm32f303xd/xe devices

russellmcc:
The 303xd and 303xe series devices do not have AF14 for pins PA11 and
PA12, so the existing implementation will not work for those devices.

According to an ST document [1], stm32f303xd and xe do not require any
particular mode for the PA11 and PA12 pin, so I've parameterized the
`usb::Peripheral` type using traits so that any mode will be
supported.  I've tried to follow a familiar pattern reminescent of the
gpio alternate function configurations.

This is of course a breaking change so I'm open to other approaches if
anyone has opinions.  This seemed to me to be the most elegant and
accurate way to handle this.

[1]: https://www.st.com/resource/en/application_note/dm00260340-migrating-between-stm32f303-and-stm32f302-line-products-stmicroelectronics.pdf

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   `serial::config::Config`. ([#239])
 - Implement `Serial::join` which allows to re-create the serial peripheral,
   when `Serial::split` was previously called. ([#252])
+- Parameterized `usb::Peripheral` and `usb::UsbType` on the pin configuration
+  used ([#254])
 
 ## [v0.7.0] - 2021-06-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,8 +20,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Breaking Changes
 
 - Refactor CAN to use the [`bxCan`](https://github.com/stm32-rs/bxcan) crate. ([#207])
-- Add support for configuring parity and stop bits in addition to baud rate for `Serial` with
-  `serial::config::Config`. ([#239])
+- Add support for configuring parity and stop bits in addition to baud rate for
+ `Serial` with `serial::config::Config`. ([#239])
 - Implement `Serial::join` which allows to re-create the serial peripheral,
   when `Serial::split` was previously called. ([#252])
 - Parameterized `usb::Peripheral` and `usb::UsbType` on the pin configuration
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Implement `Serial::join` which allows to re-create the serial peripheral,
   when `Serial::split` was previously called. ([#252])
 - Parameterized `usb::Peripheral` and `usb::UsbType` on the pin configuration
-  used ([#254])
+  used ([#255])
 
 ## [v0.7.0] - 2021-06-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -351,6 +351,7 @@ let clocks = rcc
 [defmt]: https://github.com/knurling-rs/defmt
 [filter]: https://defmt.ferrous-systems.com/filtering.html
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Refactor CAN to use the [`bxCan`](https://github.com/stm32-rs/bxcan) crate. ([#207])
 - Implement `Serial::join` which allows to re-create the serial peripheral,
   when `Serial::split` was previously called. ([#252])
+- Parameterized `usb::Peripheral` and `usb::UsbType` on the pin configuration
+  used ([#254])
```
Sh3Rm4n, line 26:
```suggestion
  used ([#255])
```

and lower in the file you have to add the line:

```markdown
[#255]: https://github.com/stm32-rs/stm32f3xx-hal/pull/255
```

to make this link work.

src/usb.rs
```diff
@@ -15,22 +15,40 @@ use crate::gpio::gpioa::{PA11, PA12};
 use crate::gpio::{PushPull, AF14};
 pub use stm32_usbd::UsbBus;
 
+/// Trait implemented by all pins that can be the "D-" pin for the USB peripheral
+pub trait DMPin: crate::private::Sealed {}
```
Sh3Rm4n, line 19:
```suggestion
pub trait DmPin: crate::private::Sealed {}
```

Because of https://github.com/stm32-rs/stm32f3xx-hal/blob/2da75a7dff397057ffe0bd82abc690c9a88af3b9/src/lib.rs#L89 this is allowed, but let's try to stick to CamelCase where possible. This also applies to `DM` -> `Dm` later in the file.

src/usb.rs
```diff
@@ -68,4 +86,5 @@ unsafe impl UsbPeripheral for Peripheral {
 ///
 /// As this MCU family has only USB peripheral,
 /// this is the only possible concrete type construction.
-pub type UsbBusType = UsbBus<Peripheral>;
+pub type UsbBusType<DP = PA11<AF14<PushPull>>, DM = PA12<AF14<PushPull>>> =
+    UsbBus<Peripheral<DM, DP>>;
```
Sh3Rm4n, line 90: Using this type for `stm32f303xd` devices with the set defaults is not possible / leads to a compiler error, am I understanding that correctly? What about providing different defaults for these devices or a different type alias?

Sh3Rm4n (changes requested):
Nice catch. Thank you! 

Implementation wise, this looks pretty reasonable, so let's go for it.

src/usb.rs
russellmcc: Thanks!

src/usb.rs
russellmcc: I've made a separate definition for the d/e devices.  It won't lead directly to a type error since `PA12<AF14<PushPull>>>` is still a valid type on the d/e devices, but it would be impossible to construct one :-)

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Refactor CAN to use the [`bxCan`](https://github.com/stm32-rs/bxcan) crate. ([#207])
 - Implement `Serial::join` which allows to re-create the serial peripheral,
   when `Serial::split` was previously called. ([#252])
+- Parameterized `usb::Peripheral` and `usb::UsbType` on the pin configuration
+  used ([#255])
```
Sh3Rm4n, line 26: There is still 

```
[#255]: https://github.com/stm32-rs/stm32f3xx-hal/pull/255
```

missing at the end of the file, to make this link work. (Line 350 onwards) 

Sorry :see_no_evil: I probably should add a comment to the changelog to make more obvious.

Sh3Rm4n (changes requested):
Sorry to bother you with the changelog. Otherwise, this looks good to go :)

russellmcc:
> Sorry to bother you with the changelog. Otherwise, this looks good to go :)

Woops!  No problem, thanks for all of your awesome work on this repo!

Sh3Rm4n (approved):
Thanks for your contribution! :+1:

PR #256, opened 2021-07-15
Enable `feature(doc_cfg)` if `cfg(docsrs)`

WaffleLapkin:
This allows for `#[cfg_attr(docsrs, doc(cfg(...)))]` to properly work.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -91,6 +91,7 @@
 #![warn(missing_docs)]
 #![deny(macro_use_extern_crate)]
 #![cfg_attr(nightly, deny(rustdoc::broken_intra_doc_links))]
+#![cfg_attr(docsrs, feature(doc_cfg))]
 
 use cfg_if::cfg_if;
 


Sh3Rm4n:
Oh wow, that is what I was missing 👀 Thanks! 👍

Just curious, how did you check it is working locally? Obviously, my method wasn't fool-proof.

WaffleLapkin:
@Sh3Rm4n I ran `cargo doc` with the `docsrs` cfg. Like so:

```console
RUSTDOCFLAGS="--cfg docsrs" cargo doc --features "stm32f303xc" --open
```

Or, rather I have the following in `~/.cargo/config`:
```toml
[build]
rustdocflags = ["--cfg", "docsrs"]
```
and used 
```console
cargo doc --features "stm32f303xc" --open
```

---

Without this patch, these commands lead to the following error:
```text
error[E0658]: `#[doc(cfg)]` is experimental
   --> src/lib.rs:130:20
    |
130 | #[cfg_attr(docsrs, doc(cfg(feature = "stm32f303")))]
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #43781 <https://github.com/rust-lang/rust/issues/43781> for more information
    = help: add `#![feature(doc_cfg)]` to the crate attributes to enable
```
This is how I actually noticed the problem while following the [Discovery guide](https://docs.rust-embedded.org/discovery/05-led-roulette/the-challenge.html).

Sh3Rm4n:
Thank you very much! I could reproduce with your change that

```bash
RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --features "stm32f303xc can stm32-usbd" --open
```

builds and also shows the feature gates. 

I'm still not sure, if docrs will build with nightly, so that this is visible on the next release. I'll have to look up that.

WaffleLapkin:
docs.rs use nightly `rustdoc`, see https://docs.rs/about/builds:
> All crates are built in a sandbox using the nightly release of the Rust compiler. The current version in use is `rustc 1.55.0-nightly (b1f8e27b7 2021-07-15)`.

Issue #236, opened 2021-06-08
Re-export each pin type aliases (PA0, PA1, ..) under gpio module


Piroro-hs:
Pros:
* less typings
* remove redundancy of *gpio*::*gpio***a**::P**A**0

Cons:
* Super long "Re-exports" section in rustdoc

Piroro-hs:
<details>
<img src="https://user-images.githubusercontent.com/22536104/121206269-49ee7e80-c8b3-11eb-899e-7e4e251fce57.png"></img>
</details>

Looks not bad😃.
However, should we keep `gpio::gpioX::PXn` for compatibility?

Sh3Rm4n:
Hm. 🤔 I don't know? Maybe because this is subtle difference to the other implementation? But maybe not? It is definitely a big change for everyone upgrading a version. 

What about a slight conversation phase? Keep them in the old place for one version while introducing the new place and in the version after that remove them? 

After I'm finally about to release 0.7 I'll hope to increase the release cadence.

PR #257, opened 2021-07-19
Define pin type aliases directly under gpio module

Piroro-hs:
#236


Piroro-hs:
Oops

Sh3Rm4n (approved):
I like your solution with the re-exports into the submodules. That makes it very clean in the generated doc :+1:

Issue #109, opened 2020-06-21
Serial Communication Issue with crate


talalahsan01:
Hey!
Basically i was building a project which uses pwm and serial communication "usart1". So the Issue i was facing is that if I use the Serial Communication from Discovery Book, it works fine but if i write the same code with stm32f3xx_hal crate , it does work but it sends data in some other encoding, for example, from discovery book example, If i press "enter" it sends "13" but by using this crate "ENTER" = "128". Also no matter what i send i receive two integers. I am  beginner so i am not very sure how to deal with this.

strom-und-spiele:
hi @talalahsan01,

if you post the code you are working with e.g. to your github account, we could have a look into it.
if you describe your set up we have a chance in understanding it.
For example I have no clue how your keyboard input is connected to your chip, how you are receiving anything, how you are displaying it, what the expected behaviour is, nor what all of this has to do with a pwm.

talalahsan01:
> hi @talalahsan01,
> 
> if you post the code you are working with e.g. to your github account, we could have a look into it.
> if you describe your set up we have a chance in understanding it.
> For example I have no clue how your keyboard input is connected to your chip, how you are receiving anything, how you are displaying it, what the expected behaviour is, nor what all of this has to do with a pwm.

I am showing here the code first of only usart1 communication that i am using
`let cp = cortex_m::Peripherals::take().unwrap();
    let dp = stm32::Peripherals::take().unwrap();

    let mut flash = dp.FLASH.constrain();
    let mut rcc = dp.RCC.constrain();

    let clocks = rcc.cfgr.freeze(&mut flash.acr);

    let mut gpioa = dp.GPIOA.split(&mut rcc.ahb);
    let tx = gpioa.pa9.into_af7(&mut gpioa.moder, &mut gpioa.afrh);
    let rx = gpioa.pa10.into_af7(&mut gpioa.moder, &mut gpioa.afrh);

    Serial::usart1(dp.USART1, (tx, rx), 9600.bps(), clocks, &mut rcc.apb2);

    let usart =  usart();

    let mut itm = cp.ITM;

    let mut buffer: Vec<u8, consts::U32> = Vec::new();

    

    loop{
        buffer.clear();

        loop{
            while usart.isr.read().rxne().bit_is_clear() {}
            let byte = usart.rdr.read().rdr().bits() as u8;
            iprintln!(&mut itm.stim[0],"{:?}", byte);
            //let ch = char::from(byte);

[...]

PR #260, opened 2021-07-28
USART1 connects to PCLK1.

richard7770:
The clock tree figures of RM0366, RM0365, RM0364, RM0316, and RM0313 show that the default setting of RCC::CFGR3::USART1SW will, for some variants, make the port use PCLK1 for reference, and not PCLK2 where port is resident.

The affected variants are stm32f3xxx4/6/8, except stm32f37xxx.

This is consistent with my observations on stm32f344r8, which fails to produce the correct baud rate:
- If sysclk is configured above 32 MHz and both pclk are untouched (pclk1 is then divided, but pclk2 is not), or
- If the pclk's are specifically configured differently.

This might be related to #109, although I don't see why the discussed code snippet should cause different pclk's.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   features, can no longer be disabled. ([#259])
   1. The "unproven" features are no longer unproven and used anywhere anyways.
   2. This crate was not building successfully without the unproven feature.
+- Set the correct baud rate for chips where `USART1SW_A::PCLK` leads to a baud rate 
+  derived from PCLK1, rather than the ports own bus clock, PCLK2. ([#260])
 
 ### Breaking Changes
 
@@ -367,6 +369,7 @@ let clocks = rcc
 [defmt]: https://github.com/knurling-rs/defmt
 [filter]: https://defmt.ferrous-systems.com/filtering.html
 
+[#260]: https://github.com/stm32-rs/stm32f3xx-hal/pull/260
 [#259]: https://github.com/stm32-rs/stm32f3xx-hal/pull/259
 [#257]: https://github.com/stm32-rs/stm32f3xx-hal/pull/257
 [#255]: https://github.com/stm32-rs/stm32f3xx-hal/pull/255
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,8 +32,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   features, can no longer be disabled. ([#259])
   1. The "unproven" features are no longer unproven and used anywhere anyways.
   2. This crate was not building successfully without the unproven feature.
-- Set the correct baud rate for chips where `USART1SW_A::PCLK` leads to a baud rate 
-  derived from PCLK1, rather than the ports own bus clock, PCLK2. ([#260])
+- Set the correct baud rate for chips where `USART1SW_A::PCLK` leads to a
+  baud rate derived from PCLK1, rather than the ports own bus clock, PCLK2.
+  ([#260])
 
 ### Breaking Changes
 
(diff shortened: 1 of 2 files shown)


Sh3Rm4n:
> This is consistent with my observations on stm32f344r8, which fails to produce the correct baud rate:
> 
> * If sysclk is configured above 32 MHz and both pclk are untouched (pclk1 is then divided, but pclk2 is not), or
> * If the pclk's are specifically configured differently.

Thanks! I'll take a deeper look later today :)

src/serial.rs
```diff
@@ -869,8 +869,17 @@ macro_rules! usart_var_clock {
 }
 
 cfg_if::cfg_if! {
-    if #[cfg(any(feature = "svd-f301", feature = "svd-f3x4"))] {
-        usart_var_clock!([(1,2)]);
+    if #[cfg(all(
+        any(
+            feature = "mem-4",
+            feature = "mem-6",
+            feature = "mem-8",
```
Sh3Rm4n, line 876: Is this feature constraint not describable via the `svd-*` features? I mean the `mem-*` features do work, but are rather intended to distinguish memory differences and this seems not related.

richard7770: I used them because, for example f302x6/8 are affected, while f302xb/c/d/e are not.

Sh3Rm4n: Hm. Yeah, this is not representable with the `svd-*`-features. But I would suggest to used `stm32f303xb`, `stm32f303xc` ... so that it is more obvious, when reading it :)

Sh3Rm4n: > (Fig 12+13 vs. Fig 14 of RM0365)

Thanks for that info! I can see the difference. So I think the implementation is fine. Will still look at it later today :)

Sh3Rm4n:
As this is a bugfix effecting `v0.7` would you like to add a CHANGELOG entry?

richard7770:
I didn't mention #109 in the changelog, because I'm unsure whether this is a fix to that bug.

Sh3Rm4n (approved):

Sh3Rm4n:
> I didn't mention #109 in the changelog, because I'm unsure whether this is a fix to that bug.

That's okay. I'll have to invest a bit more time on that issue anyway, and as it might not be related, let's ignore it for now.

CHANGELOG.md
```diff
@@ -32,6 +32,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   features, can no longer be disabled. ([#259])
   1. The "unproven" features are no longer unproven and used anywhere anyways.
   2. This crate was not building successfully without the unproven feature.
+- Set the correct baud rate for chips where `USART1SW_A::PCLK` leads to a baud rate 
+  derived from PCLK1, rather than the ports own bus clock, PCLK2. ([#260])
```
Sh3Rm4n, line 36:
```suggestion
- Set the correct baud rate for chips where `USART1SW_A::PCLK` leads to a
  baud rate derived from PCLK1, rather than the ports own bus clock, PCLK2. 
  ([#260])
```

Markdown linter is unhappy :( https://github.com/stm32-rs/stm32f3xx-hal/pull/260/checks?check_run_id=3182875827

Sh3Rm4n:
> If sysclk is configured above 32 MHz and both pclk are untouched (pclk1 is then divided, but pclk2 is not), or

That reminds me, that I probably should change the processor clock to 32 MHz or above instead of the default 8 Mhz for the [uart tests](https://github.com/stm32-rs/stm32f3xx-hal/blob/ee27f328639c3c75c6f5ab61be06dbccea883df4/testsuite/tests/uart.rs#L64)

Sh3Rm4n:
I have not run a local test to verify this behavior nor check the erroneous old behavior, but I'll trust you and the datasheet on that one. 

Thanks for the catch and this contribution. Much appreciated! :+1:

PR #267, opened 2021-08-01
Rework Timer

Sh3Rm4n:


src/timer.rs
```diff
@@ -147,424 +124,540 @@ pub enum Event {
...
+    where
+        T: Into<Self::Time>,
+    {
+        self.stop();
 
-                    self
-                }
+        let frequency = timeout.into().integer();
+        let timer_clock = TIM::clock(&self.clocks);
+        // TODO: Obtain ppre1 and pclk1 (is ppre1 even the correct ppre???)
```
Piroro-hs, line 273: > is ppre1 even the correct ppre???

TIM peripherals on APB1 uses ppre1 and on APB2 uses ppre2

Sh3Rm4n: Thanks for the quick lookup. I suspected something like this.

src/serial.rs
```diff
@@ -722,9 +722,10 @@ where
         self.usart.isr.read().busy().bit_is_set()
     }
 
-    /// Obtain the assocated intterupt number for the serial peripheral.
+    /// Obtain the associated interupt number for the serial peripheral.
```
Sh3Rm4n, line 725:
```suggestion
    /// Obtain the associated interrupt number for the serial peripheral.
```

PR #279, opened 2021-08-16
Set correct cargo release push remote

Sh3Rm4n:
- [x] Add comment that `upstream` is used instead of `origin`.

### DIFF
--- a/release.toml
+++ b/release.toml
@@ -2,7 +2,8 @@
 
 tag-message = "{{version}}"
 no-dev-version = true
-pre-release-commit-message = "Release {{version}}"
+pre-release-commit-message = "Release v{{version}}"
+push-remote = "upstream"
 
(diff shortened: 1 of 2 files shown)


release.toml
```diff
@@ -3,6 +3,7 @@
 tag-message = "{{version}}"
 no-dev-version = true
 pre-release-commit-message = "Release {{version}}"
+push-remote = upstream
```
Sh3Rm4n, line 6:
```suggestion
push-remote = "upstream"
```

release.toml
```diff
@@ -3,6 +3,7 @@
 tag-message = "{{version}}"
 no-dev-version = true
 pre-release-commit-message = "Release {{version}}"
```
Sh3Rm4n, line 5:
```suggestion
pre-release-commit-message = "Release v{{version}}"
```

PR #283, opened 2021-08-19
Set some features as the default

Sh3Rm4n:


Cargo.toml
```diff
@@ -35,23 +35,22 @@ rustc-args = ["--cfg", "docsrs"]
 [dependencies]
 cfg-if = "1.0.0"
 cortex-m = "0.7.2"
-cortex-m-rt = "0.6.14"
+cortex-m-rt = { version = "0.7.0", features = ["device"] }
```
adamgreig, line 38: Since you're making `rt` a default feature, and it enables `stm32f3/rt`, which enables `cortex-m-rt/device`, I don't know if there's any point enabling `cortex-m-rt/device` here: if `cortex-m-rt/device` is enabled but `stm32f3/rt` is not enabled, I don't think the build would work properly at all. I might be missing something you've thought of though.

I'm all in favour of making `rt` default (in fact the next version of stm32f3 will have `rt` as a default feature anyway), just not sure `cortex-m-rt/device` should be enabled separately.

Sh3Rm4n: > I might be missing something you've thought of though.

I have not given it much thought at all 😅, well I had a slight hunch that this might be superfluous. So thanks for pointing that out.

PR #287, opened 2021-09-06
examples/serial_echo_rtic.rs: add more explanations; usage

Lotterleben:
I found the `serial_echo_rtic.rs` example to be not completely self-explanatory. This PR attempts to add info that I would've found helpful while reading/using it :)

### DIFF
--- a/examples/serial_echo_rtic.rs
+++ b/examples/serial_echo_rtic.rs
@@ -3,6 +3,14 @@
 
 use panic_rtt_target as _;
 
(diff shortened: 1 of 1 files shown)


Lotterleben:
whoops, please ignore the close + reopen

Sh3Rm4n:
Thank you very much. More documentation is always welcome! 

In the examples, some concepts are used, which are not immediately obvious. Having worked with these controllers for a while I sometimes forget, which concepts need some more explanation 😅

PR #300, opened 2021-11-16
update cortex-m-rt, defmt, bxcan

burrbull:


Sh3Rm4n (approved):

Cargo.toml
```diff
@@ -36,26 +36,26 @@ rustc-args = ["--cfg", "docsrs"]
 [dependencies]
 cfg-if = "1.0.0"
 cortex-m = "0.7.2"
-cortex-m-rt = { version = ">=0.6.13, <0.8" }
-defmt = { version = "0.2.3", optional = true }
+cortex-m-rt = "0.7"
+defmt = { version = "0.3.0", optional = true }
```
Sh3Rm4n, line 40:
```suggestion
defmt = { version = ">=0.2.3, <0.4.0", optional = true }
```
As this does not require any changes in the code, couldn't we also still support `0.2.3`?

Sh3Rm4n (changes requested):

Sh3Rm4n (approved):

PR #309, opened 2022-02-08
Re license project to MIT / APACHE

Sh3Rm4n:
Disclaimer: IANAL

~~This crate using 0BSD before was violating licenses of the other
dependencies, which use MIT or APACHE for the most part.~~

~~These licenses are stricter than 0BSD, so this project was not
allowed to use these dependencies before IIUC.~~

~~To solve this issue,~~ re-license the project to be under MIT / APACHE.

This should be possible without problem, as with 0BSD it is not needed
to ask previous contributors about the license change and is compatible
with MIT / APACHE.


adamgreig:
I'm also not a lawyer and of course you're welcome to relicence to MIT/Apache, but I don't think you need to - it's not a problem to use MIT/Apache licensed dependencies in a 0BSD licensed crate, so long as any eventual distribution follows the terms of those other licences. That's what it means for them to be compatible licences. Plus, stm32f3xx-hal doesn't even distribute code from those other projects, it just requires the user provide it at build-time -- really the obligation is on anyone distributing a composition of this library and its dependencies (like, an application binary) to ensure all the licences are obeyed, which basically just means including the relevant copyright headers.

Sh3Rm4n:
> it's not a problem to use MIT/Apache licensed dependencies in a 0BSD licensed crate, so long as any eventual distribution follows the terms of those other licences.

Thanks for this information. Yeah, that makes sense, I haven't even considered that. And good to know that it is not a license violation to use `0BSD` for this crate while using `MIT / APACHE` for its dependencies.

Anyway, another motivation of mine is the following: I've thought for a long while to match the licensing to the rest of the ecosystem (all of which are MIT), so even when someone uses this crate it has to follow the terms of condition of the dependencies anyway (which are the MIT / Apache conditions). So `0BSD` isn't even that useful to use for this particular crate it seems to me.

PR #316, opened 2022-03-30
f4xx-hal like gpio

burrbull:


CHANGELOG.md
```diff
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
-No changes.
+- Use const generics for pins. The MSRV was bumped to 1.59 ([#316])
```
Sh3Rm4n, line 17: As this pull request changes many things in the gpio API I'd like to have a short list of things which changed like:

* `into_af_push_pull` -> `into_alternate`
* `downgrade().downgrade()` -> `erase()`

Kind of like a short upgrade guide. You don't have to be too detailed. I'm fine to do it myself as well. You can also wait with that change until I finished my thorough review :)

Sh3Rm4n, line 17: As this pull request changes many things in the gpio API I'd like to have a short list of things which changed like:

* `into_af_push_pull` -> `into_alternate`
* `downgrade().downgrade()` -> `erase()`

Kind of like a short upgrade guide. You don't have to be too detailed. I'm fine to do it myself as well. You can also wait with that change until I finished my thorough review :) 

Also, this should be listed under `### Breaking Changes`

src/gpio.rs
```diff
@@ -59,498 +63,197 @@
...
+//! The above mode change methods guarantee that you can only call input functions when the pin is
+//! in input mode, and output when in output modes, but can lead to some issues. Therefore, there
+//! is also a mode where the state is kept track of at runtime, allowing you to change the mode
+//! often, and without problems with ownership, or references, at the cost of some performance and
+//! the risk of runtime errors.
+//!
+//! To make a pin dynamic, use the `into_dynamic` function, and then use the `make_<mode>` functions to
+//! change the mode
+
+#![allow(missing_docs)]
```
Sh3Rm4n, line 81: Please remove this when finished. If it does make sense to not comment, some parts either add this attributes to local functions or types or use `#[doc(hidden)]` in case of traits with the private method convention e.g. the functions or prefixed with `fn _`, which (I like by the way).

src/gpio.rs
```diff
@@ -59,498 +63,197 @@
...
-use core::{convert::Infallible, marker::PhantomData};
+mod convert;
+use convert::PinMode;
+mod partially_erased;
+pub use partially_erased::{PEPin, PartiallyErasedPin};
+mod erased;
+pub use erased::{EPin, ErasedPin};
+mod dynamic;
+pub use dynamic::{Dynamic, DynamicPin};
+mod hal_02;
```
Sh3Rm4n, line 97: Hm, `hal_02` seem a little cryptic for a module name. Maybe use the conventional abbreviation of `embeded_hal` like `eh`

Or as it is not typed very often I'm fine with a long name like `embedded_hal_v0_2` or similar.

src/gpio.rs
```diff
@@ -59,498 +63,197 @@
...
-    pub trait Index {
-        #[doc(hidden)]
-        fn index(&self) -> u8;
-    }
+/// Analog mode (type state)
+pub struct Analog;
+
+pub type Debugger = Alternate<0, PushPull>;
 
+mod sealed {
```
Sh3Rm4n, line 159: This could probably still be named `marker` though I'm fine with that change.

src/gpio.rs
```diff
@@ -59,498 +63,197 @@
...
-    /// Pulled up
-    PullUp,
-    /// Pulled down
-    PullDown,
+    Low = 0,
+    Medium = 1,
+    High = 3,
 }
 
-/// GPIO interrupt trigger edge selection
```
Sh3Rm4n, line 293: Why are these doc comments removed?

src/gpio.rs
```diff
@@ -59,498 +63,197 @@
...
-    fn is_high(&self) -> Result<bool, Self::Error> {
-        Ok(!self.is_low()?)
-    }
-
-    fn is_low(&self) -> Result<bool, Self::Error> {
-        // NOTE(unsafe) atomic read with no side effects
-        Ok(unsafe { (*self.gpio.ptr()).is_low(self.index.index()) })
-    }
-}
+cr!(false: [0, 1, 2, 3, 4, 5, 6, 7]);
```
Sh3Rm4n, line 238: What does cr mean?

src/gpio.rs
```diff
@@ -653,957 +352,833 @@ where
...
+    /// a short spike of an incorrect value
+    #[inline(always)]
+    fn _set_state(&mut self, state: PinState) {
+        match state {
+            PinState::High => self._set_high(),
+            PinState::Low => self._set_low(),
         }
+    }
+    #[inline(always)]
+    fn _set_high(&mut self) {
```
Sh3Rm4n, line 503: If this function is private, adding `#[doc(hidden)]` might make sense.

src/gpio.rs
```diff
@@ -653,957 +352,833 @@ where
...
-                ($GPIOX, $gpioy::moder::MODER15_A, 2);
-                impl Moder for MODER {
-                    fn input { INPUT }
-                    fn output { OUTPUT }
-                    fn alternate { ALTERNATE }
-                    fn analog { ANALOG }
-                }
-            }
+#[cfg(feature = "gpio-f302")]
+gpio!(GPIOA, gpioa, PA, 'A', PAn, [
```
Sh3Rm4n, line 655: The macros here are different to the previous macros, which were auto generated by the codegen tools inside this repo. 

Maybe you can adjust that tool to get your desired output. But I'm fine if you don't like to bother. `codegen` needs a cleanup / overhaul anyway. 

But did you autogenerate this or did you adjust these invocations by hand, basing on the previous implementation?

src/gpio/erased.rs
```diff
@@ -0,0 +1,147 @@
+use super::*;
+
+pub type EPin<MODE> = ErasedPin<MODE>;
```
Sh3Rm4n, line 3: I don't think this type alias is needed. I think `ErasedPin` is short enough. At least this type alias should not be used in this code base to benefit readability.

src/gpio/partially_erased.rs
```diff
@@ -0,0 +1,129 @@
+use super::*;
+
+pub type PEPin<const P: char, MODE> = PartiallyErasedPin<P, MODE>;
```
Sh3Rm4n, line 3: As the other type alias, I'm not a fan of this one. Though here the case is stronger for an alias as the original type name is long. 

But I don't think this is a problem. A user might just use a type alias themself via `use hal::gpio::PartiallyErasedPin as PEPin`.

src/prelude.rs
```diff
@@ -16,9 +16,3 @@ pub use crate::rcc::RccExt as _stm32f3xx_hal_rcc_RccExt;
 pub use crate::syscfg::SysCfgExt as _stm32f3xx_hal_syscfg_SysCfgExt;
 pub use crate::time::duration::Extensions as _stm32f3xx_hal_time_time_Extensions;
 pub use crate::time::rate::Extensions as _stm32f3xx_hal_time_rate_Extensions;
-pub use crate::{
```
Sh3Rm4n, line 19: This is an important breaking change, which should be commented especially in the Changelog. 

Might also be a good idea to put this in a single commit :)

Sh3Rm4n (commented):
Wow, thank you very much for this effort and to keep the API  aligned between `f3` and `f4`. ❤️ 

This change is rather big. I'll probably have time to thoroughly review it this weekend.

CHANGELOG.md
burrbull: I can add functions with old names and mark them `deprecated`.

Sh3Rm4n: Probably it is not really worth it to add deprecation notices as the resulting code would be really messy I guess.

src/lib.rs
```diff
@@ -300,3 +300,9 @@ impl From<bool> for Toggle {
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub struct TryFromIntError;
+
+fn stripped_type_name<T>() -> &'static str {
+    let s = core::any::type_name::<T>();
+    let p = s.split("::");
+    p.last().unwrap()
```
Sh3Rm4n, line 307: To safe one unwrap call, this could be done, I guess:

```suggestion
    p.last().unwrap_or(s)
```

Sh3Rm4n:
Hey, sorry for the delay, I'm pretty busy right now. I hope I'll make progress on this review this week.

PR #318, opened 2022-06-14
Minimal DAC driver

jonas-schievink:
This PR adds a minimal DAC driver that supports manual sample output on channel 1 only, and does not manage pins. I hope that's okay and you'll still consider merging this :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,12 +14,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Added
+
+- Add a minimal DAC driver ([#318])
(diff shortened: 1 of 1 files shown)


jonas-schievink:
Friendly ping @Sh3Rm4n :)

Sh3Rm4n:
Thanks! That's very much a minimal implementation. Wasn't aware that the implementation is so easy :grinning: 

Sorry for the delay. With the addition of a CHANGELOG entry, this would be good to go!

Sh3Rm4n (approved):

jonas-schievink:
Thanks for the review! I've added a changelog entry.

PR #326, opened 2022-12-28
Update RTC API

robamu:
The functions [`from_ymd`](https://docs.rs/chrono/latest/chrono/naive/struct.NaiveDate.html#method.from_ymd) and [`and_hms`](https://docs.rs/chrono/latest/chrono/naive/struct.NaiveDate.html#method.and_hms) were deprecated in 0.4.23 in favour of the `_opt` variants


Sh3Rm4n:
Thanks! :) I've added the liberty to add some error handling instead of unwrapping.

Issue #325, opened 2022-12-26
Using Serial DMA for variable sized transfers


robamu:
Hello,

I think this question might also be suited in a forum post, but considering I am using a STM32F3-Disco in my current project, I think its okay to ask it here.

I am currently planning on how to perform variably sized transfers using DMA + IRQ + RTIC. According to some forum posts, this is definitely possible using the IDLE line or the Receiver Timeout IRQ by the USART peripheral. More specifically, I wanted to have one RTIC task which uses DMA to send all TX data, and one RTIC task to receive variable sized RX packets, which are only limited in their maximum size.

Unfortunately, I have not really found an example for this in Rust yet and I only have experience with variable sized transfers using a combination of [FIFO + IRQ on a smaller Cortex-M0 based system](https://docs.rs/va108xx-hal/latest/va108xx_hal/uart/struct.UartWithIrqBase.html#method.read_fixed_len_using_irq). Most of the DMA based API I have found uses the `read_exact` API which uses a fixed buffer size.

I've looked around and experimented myself a little bit. I have found this code snippet: https://github.com/kalkyl/f303-rtic/blob/main/src/bin/serial.rs , which goes into the direction of what I require.

Some thoughts I had:

1. TX: is usually a little bit easier. I have a RTIC task which reads from a heapless queue or some other form of queue and sends any contained frames with DMA. My first approach would be to simply wait until the transfer is complete because I'd like to unblock the shared serial as soon as possible. The second approach instead uses a transfer complete IRQ, but then I am not fully sure how to access the USART peripheral without trickery or hacks just to read IRQ events, for example for pending RX
2. RX: I configure a receive timeout IRQ and then initiate a DMA transfer with the maximum expected packet length. When receiving a  recv timeout IRQ or a transfer complete, I stop the DMA transfer (for recv timeout) and restart a new one with the maximum expected packet length. I pass the read packet into a heapless queue or some other form of queue which is processed by another task.

Now, from what I have see from the STM32F3 HAL docs, this will not  be straightforward.
[...]

PR #328, opened 2022-12-29
UART DMA improvements and refactoring

robamu:
Some preparation to allow variable sized DMA transfers as specified in https://github.com/stm32-rs/stm32f3xx-hal/issues/325 . I did not have to change anything in the `serial_dma` example, but this is unfortunately a breaking change. In the process of reading a it about DMA, I also saw that the datasheet recommends checking the UART TC flag for transmission completeness. I am not sure whether this was not done on purpose or simply overlooked, but the changes ensure the TC flag is now checked as well.

Copied from CHANGELOG:

### Changed

- serial: The DMA functions `write_all` and `read_exact` now returns
  the wrapper structs `SerialDmaTx` and `SerialDmaRx` instead of the direct
  DMA transfer struct. These allow checking the USART ISR events
  with `is_event_triggered` as well.

### Fixed

- serial: The previous DMA `write_all` implementation did use the DMA transfer completion
  event to check for transfer completion, but MCU datasheet specifies that the TC
  flag of the USART peripheral should be checked for transfer completion to avoid
  corruption of the last transfer. This is now done by the new `SerialDmaTx` wrapper.

## Added

- serial: Public `is_event_triggered` method which allows to check for events
  given an event and a USART reference.

If you consider this mergeable (with adaptions/changes), I also have some other branches which build on top of these changes, for example the splitting of RX and TX events.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,25 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 No changes.
 
+### Changed
+
+- serial: The DMA functions `write_all` and `read_exact` now returns
+  the wrapper structs `SerialDmaTx` and `SerialDmaRx` instead of the direct
+  DMA transfer struct. These allow checking the USART ISR events
+  with `is_event_triggered` as well.
+
+### Fixed
+
+- serial: The previous DMA `write_all` implementation did use the DMA transfer completion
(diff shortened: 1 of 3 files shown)


robamu:
Looking at this again as I am updating another application. This probably needs to be re-worked now that the `split` has been removed. Then again, using DMA requires TX and RX to be different entities.

PR #330, opened 2023-01-01
Minor clippy and deprecation fixes

barafael:


Sh3Rm4n:
Thanks! :)

I've used a different approach for the rtc update, but I'll take the Cargo.toml cleanup

PR #343, opened 2023-09-04
extend ticks calculation in timer start from u32 to u64

### LINKED ISSUE #342: Integer saturation in Timer::start
The following code snippets should be equivalent in result, but they are not:

```rust
let mut timer = Timer::new(cx.device.TIM2, clocks, &mut rcc.apb1);
timer.enable_interrupt(Event::Update);
timer.start(1.milliseconds());
```

```rust
let mut timer = Timer::new(cx.device.TIM2, clocks, &mut rcc.apb1);
timer.enable_interrupt(Event::Update);
timer.start(1000.microseconds());
```

This happens due to the integer saturation in following line:
https://github.com/stm32-rs/stm32f3xx-hal/blob/8933e1191ea9a52a61ccf67cd8f71583b1876bad/src/timer.rs#L280

Yes, the simplest answer would be "don't do that!", but this also happens with legitimate examples like `100.microseconds()`.

Perhaps one of the following fixes can be applied:
- `ticks` can be extended to u64 (more precise result)
- change the multiplication order as follows (better performance and code size?)
```rust
let ticks = ((clock.integer() * *timeout.scaling_factor()) as u32).saturating_mul(timeout.integer());
```

apatrushev:
Fixes #342

### DIFF
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -279,9 +279,9 @@ where
         let timeout: Self::Time = timeout.into();
         let clock = TIM::clock(&self.clocks);
 
-        let ticks = clock.integer().saturating_mul(timeout.integer()) * *timeout.scaling_factor();
+        let ticks = (clock.integer() as u64).saturating_mul(timeout.integer() as u64) * *timeout.scaling_factor();
 
-        let psc: u32 = (ticks.saturating_sub(1)) / (1 << 16);
+        let psc = ticks.saturating_sub(1) / (1 << 16);
         self.tim.set_psc(crate::unwrap!(u16::try_from(psc).ok()));
 
         let mut arr = ticks / psc.saturating_add(1);


src/timer.rs
```diff
@@ -279,9 +279,9 @@ where
         let timeout: Self::Time = timeout.into();
         let clock = TIM::clock(&self.clocks);
 
-        let ticks = clock.integer().saturating_mul(timeout.integer()) * *timeout.scaling_factor();
+        let ticks = (clock.integer() as u64).saturating_mul(timeout.integer() as u64) * *timeout.scaling_factor();
```
Sh3Rm4n, line 282: This is definitely a good improvement, please use `u64::from` instead of as casts :) 

```suggestion
        let ticks = u64::from(clock.integer()).saturating_mul(u64::from(timeout.integer()) * *timeout.scaling_factor();
```

Sh3Rm4n (changes requested):

Sh3Rm4n:
Thanks, I've rebased and merged it in https://github.com/stm32-rs/stm32f3xx-hal/pull/356

PR #346, opened 2023-11-21
Fix undefined behavior compile error

Sh3Rm4n:


adamgreig:
It looks like this changes the access of the data register from `Word`-sized (which I think might be u8 or u16?) to always be a 32-bit access. Annoyingly the data register is access-size-sensitive, and doing a 32-bit read can cause up to 4 8-bit words to be popped from the peripheral's FIFO. This is the same problem as https://github.com/stm32-rs/stm32f7xx-hal/pull/207

Sh3Rm4n:
Thanks for the info. This means, we currently can not read 8 or 16 bits from the data register through the svd2rust API, but casting is not the correct way either?

PR #362, opened 2024-02-23
RtcBuilder for Rtc easy initialization and access

marekjastrzebski:
Added a builder that facilitates the initialization of RTC with various clock sources (LSE or LSI, HSE in future). The intent of the update is to easy access to other clock sources without the need for detailed knowledge of this modules configuration while still offering the ability to modify default settings. The solution does not affect the existing setup so as not to force current users to update their existing code. Easy to expand.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
-No changes.
+- Added RtcBuilder for rtc module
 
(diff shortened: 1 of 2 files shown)


marekjastrzebski:
Not sure if I can do anything to fix this checks

marekjastrzebski:
@Sh3Rm4n Hi, why do we use chrono 0.4.34, it is for rustc 1.61 we should use chrono 0.4.31. Do you want me to fix it by this PR?

PR #371, opened 2024-12-04
Update to new pac

usbalbin:
This updates this hal to use the stm32f3-staging pac. I am doing this in an attempt to [get the HRTIM peripheral to work](https://github.com/usbalbin/stm32-hrtim/pull/1).

### DIFF
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -33,7 +33,7 @@ use crate::{
     Switch,
 };
 
-use crate::pac::{adc1_2, adc1_2::ccr::CKMODE_A};
+use crate::pac::{adc1_2, adc1_2::ccr::CKMODE};
 
 #[cfg(feature = "enumset")]
 use enumset::{EnumSet, EnumSetType};
@@ -273,7 +273,7 @@ macro_rules! sp_pins {
                     // (Adc<ADC1, Disabled>, Adc<ADC2, Disabled>)
                     _adcs: &mut <Common as CommonInstance>::Childs,
                 ) -> Self {
-                    common_adc.reg.ccr.modify(|_, w| w.$en().enabled());
+                    common_adc.reg.ccr().modify(|_, w| w.$en().enabled());
                     Self { _adc: PhantomData }
                 }
             }
@@ -342,14 +342,14 @@ where
     #[must_use]
     #[inline]
     pub fn data_register(&self) -> u16 {
-        self.reg.dr.read().rdata().bits()
+        self.reg.dr().read().rdata().bits()
     }
 
     /// Returns the address of the ADC data register. Primarily useful for configuring DMA.
     // TODO(Sh3Rm4n): Check against and integrate to DMA
     #[inline]
     pub fn data_register_address(&self) -> u32 {
-        core::ptr::addr_of!(self.reg.dr) as u32
+        self.reg.dr().as_ptr() as u32
     }
 
     /// Manually start a conversion sequence.
@@ -358,7 +358,7 @@ where
     /// * The conversion mode is configured via [`Adc::set_conversion_mode`].
     #[inline]
     pub fn start_conversion(&mut self) {
-        self.reg.cr.modify(|_, w| w.adstart().start_conversion());
+        self.reg.cr().modify(|_, w| w.adstart().start_conversion());
     }
 
     /// Configure and convert the [`Adc`] instance into [`OneShot`] mode.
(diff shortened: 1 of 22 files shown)


usbalbin:
I had to enable the feature `critical-section` of the pac to gain access to the method `Peripherals::take()` which is used in the examples. Further, an implementation is needed or there was a linker error so I also had to enable `critical-section-single-core` for cortex-m. The last part, I believe, enabled some extra examples which I do not think the CI has been checking before. This is now the most if not all the errors I am getting in the CI now

src/dma.rs
```diff
@@ -365,7 +365,9 @@ pub trait Channel: private::Channel {
     fn set_transfer_length(&mut self, len: u16) {
         crate::assert!(!self.is_enabled());
 
-        self.ch().ndtr.write(|w| w.ndt().bits(len));
+        unsafe {
+            self.ch().ndtr().write(|w| w.ndt().bits(len));
+        }
```
burrbull, line 370:
```suggestion
        self.ch().ndtr.write(|w| w.ndt().set(len));
```

src/i2c.rs
```diff
@@ -190,21 +190,23 @@ impl<I2C, SCL, SDA> I2c<I2C, (SCL, SDA)> {
...
+                    .sdadel()
+                    .bits(crate::unwrap!(u8::try_from(sdadel)))
+                    .scldel()
+                    .bits(crate::unwrap!(u8::try_from(scldel)))
+                    .scll()
+                    .bits(scl_low)
+                    .sclh()
+                    .bits(scl_high)
+            });
+        }
```
burrbull:
```suggestion
        i2c.timingr.write(|w| {
            w.presc().set(crate::unwrap!(u8::try_from(presc)));
            w.sdadel().set(crate::unwrap!(u8::try_from(sdadel)));
            w.scldel().setcrate::unwrap!(u8::try_from(scldel)));
            w.scll().set(scl_low);
            w.sclh().set(scl_high)
        });
```

src/dma.rs
burrbull: `fn unsafe bits` -> `fn unsafe bits`
`fn bits` -> `fn set`

src/i2c.rs
usbalbin: Thanks

I have gone through all files and tried to replace a lot of the `.bits` with `.set` in c813ea1

src/watchdog.rs
```diff
@@ -125,21 +125,21 @@ impl IndependentWatchDog {
     #[must_use]
     pub fn interval(&self) -> Milliseconds {
         // If the prescaler was changed wait until the change procedure is finished.
-        while self.iwdg.sr.read().pvu().bit() {}
+        while self.iwdg.sr().read().pvu().bit() {}
 
-        let psc = self.iwdg.pr.read().pr().variant();
-        let reload = self.iwdg.rlr.read().rl().bits();
+        let psc = self.iwdg.pr().read().pr().variant().unwrap(); // <--- Unwrap!!
```
usbalbin: Any thoughts as to what to do here? Should we provide a default or something? Or can this possibly fail?

burrbull: Looks like H5 guy has broke this:
https://github.com/stm32-rs/stm32-rs/commit/c812a9e90019ab314ca761cf35215a040d987e4f

There should be default 256 value

src/dac.rs
```diff
@@ -27,12 +27,12 @@ impl Dac {
...
     /// Only the low 12 bits of `data` will be used, the rest is ignored.
     pub fn write_data(&mut self, data: u16) {
-        self.regs.dhr12r1.write(|w| {
+        self.regs.dhr12r1().write(|w| {
             #[allow(unused_unsafe)]
             // SAFETY: Direct write to register for easier sharing between different stm32f3xx svd
             // generated API
             unsafe {
-                w.dacc1dhr().bits(data)
+                w.dacc1dhr().set(data)
```
usbalbin: For some reason this seems to work fine for all(or most) but the f373

```
error[E0599]: no method named `set` found for struct `FieldWriter<'_, stm32f3_staging::stm32f373::dac1::dhr12r1::DHR12R1rs, 12, u16>` in the current scope
   --> src/dac.rs:35:30
    |
35  |                 w.dacc1dhr().set(data)
    |                 -            ^^^
```

src/watchdog.rs
usbalbin: ok

src/dac.rs
burrbull: then use `unsafe bits` until someone fix this in `stm32-rs`

src/watchdog.rs
burrbull: https://github.com/stm32-rs/stm32-rs/tree/iwdg-pr

usbalbin: Perfect, thanks a lot! https://github.com/stm32-rs/stm32-rs/pull/1045

usbalbin:
Correct me if I am wrong, but I do not think I am guilty to this one

```rust
error[E0599]: no method named `split` found for struct `Serial` in the current scope
  --> examples/serial_dma.rs:44:27
   |
44 |     let (tx, rx) = serial.split();
   |                           ^^^^^ method not found in `Serial<USART1, (Pin<Gpioa, U<9>, Alternate<PushPull, 7>>, Pin<Gpioa, U<10>, Alternate<PushPull, 7>>)>`
   |
   = note: the full type name has been written to '/home/runner/work/stm32f3xx-hal/stm32f3xx-hal/target/thumbv7em-none-eabihf/debug/examples/serial_dma-091110dd37752b66.long-type-1484845090031692940.txt'
   = note: consider using `--verbose` to print the full type name to the console
```

usbalbin:
@burrbull would you like to take a look at this?

src/spi.rs
```diff
@@ -372,7 +372,7 @@ where
         } else if sr.crcerr().is_no_match() {
             nb::Error::Other(Error::Crc)
         } else if sr.txe().is_empty() {
-            let write_ptr = core::ptr::addr_of!(self.spi.dr) as *mut Word;
+            let write_ptr = self.spi.dr().as_ptr() as *mut Word;
             // SAFETY: Write to register owned by this Spi struct
             unsafe { core::ptr::write_volatile(write_ptr, word) };
```
burrbull: I'd recommend to use `dr` register for `Word = u16` and `dr8` register for `Word = u8` so you don't need for casting:
```rust
self.spi.dr8().write(|w| dr().set(word));
```
Same for reading.

src/adc.rs
```diff
@@ -719,25 +719,25 @@ where
...
-            channel::Id::Ten => self.reg.smpr2.read().smp10().variant().into(),
-            channel::Id::Eleven => self.reg.smpr2.read().smp11().variant().into(),
-            channel::Id::Twelve => self.reg.smpr2.read().smp12().variant().into(),
-            channel::Id::Thirteen => self.reg.smpr2.read().smp13().variant().into(),
-            channel::Id::Fourteen => self.reg.smpr2.read().smp14().variant().into(),
-            channel::Id::Fifteen => self.reg.smpr2.read().smp15().variant().into(),
-            channel::Id::Sixteen => self.reg.smpr2.read().smp16().variant().into(),
-            channel::Id::Seventeen => self.reg.smpr2.read().smp17().variant().into(),
-            channel::Id::Eighteen => self.reg.smpr2.read().smp18().variant().into(),
+            channel::Id::Zero => self.adc.smpr1().read().smp0().variant().into(),
```
burrbull: If you specify `channel::Id::Zero` as `0` you could write it shorter:
```rust
let ch = channel as u8 
if ch < 9 {
    self.adc.smpr1.read().smp(ch).variant().into()
} else {
    self.adc.smpr1.read().smp(ch - 9).variant().into()
}
```

src/spi.rs
usbalbin: Something like 9cb658e?

burrbull: Yes

src/adc.rs
usbalbin: Like this a8f8835?

**Edit:** I think i got some errors

usbalbin: They dont seem to be array-ified

burrbull: Sorry. You are right. I've not finished this yet.

usbalbin: No problem :)

burrbull: Should work on [nightlies](https://github.com/stm32-rs/stm32-rs-nightlies) now.

usbalbin: Reverted my revert :)

usbalbin:
There seems to be quite a lot of problems with [rm0366](https://www.st.com/resource/en/reference_manual/rm0366-stm32f301x68-and-stm32f318x8-advanced-armbased-32bit-mcus-stmicroelectronics.pdf) devices. Again I do not think I introduced those issues in this PR. (for example they not having spi1 etc.)

